PR #9, opened 2021-02-17
notebooks and fixes to adjoint+gradient

krober10nd:
* remove is_owner from solvers 
*  begin work on notebooks


krober10nd:
gradient will be fixed tomorrow to be consistent with placing c^2 on the mass matrix (instead of the stiffness matrix). This avoids the need to save the vectorial values for the forward equation as well.

PR #21, opened 2021-04-22
New source

Olender:
Updating new source into eageLatinAmerica branch to compare grid sweep results


spyro/sources/Sources.py
```diff
@@ -19,75 +21,40 @@ def __init__(self, model, mesh, V, comm):
...
+        self.cell_tabulations = None
+        self.cellNodeMaps = None
+        self.nodes_per_cell = None
+
+    def apply_source(self, rhs_forcing,value):
+        """ Applies source in a assembled right hand side.
+        """
+        for source_id in range(self.num_receivers):
+            for i in range(len(self.cellNodeMaps[source_id])):
+                rhs_forcing.dat.data[int(self.cellNodeMaps[source_id][i])] = value * self.cell_tabulations[source_id][i]
```
krober10nd, line 54: Does this assume only one source?

Olender: No, it goes through every source with for source_id in range(self.num_receivers). It says num_receivers because itś a subclass from Receivers, but that value comes from the number of sources in the input file in the __init__.

Olender: It should work with multiple sources and parallelism since receivers works with multiple receivers and parallelism. However this still needs to be tested to make sure.

This PR was done in the testing_eage branch so I could compare it with the figures we generated for EAGE Latin AMerica to see if there really is an improvement.  I am still running the grid sweep to compare and that's going to take awhile. After that, if it really is better, I am going to test multiple sources and parallelism before doing a PR on the master branch

PR #29, opened 2021-04-28
Notebooks able to use Leapfrog functions

Ig-dolci:
I built a Leapfrof_AD (only used for automatic differentiation)


krober10nd (changes requested):
In the notebooks could we avoid import the PIL package? Also, there's a function in spyro.plot to plot the shot record already so need for graph2dec.

krober10nd (approved):

Issue #27, opened 2021-04-24
nothing leapfrog about leapfrog


krober10nd:
* need to remove the usage of the word ``leapfrog`` throughout. it's a central FD in time...

PR #31, opened 2021-04-30
renaming leapfrog to forward and leapfrog adjoint to gradient

krober10nd:
* rename leapfrog to forward and leapfrog_adjoint to gradient. #27 
* remove definition of wavelet to outside of solvers (as these should be able to be arbitrarily specified) #30


krober10nd:
* PML keyword is now called "BCs" as it is does not just control the PML

PR #36, opened 2021-05-05
support for sources and receivers in spatial parallel

krober10nd:
* Addressing simulation with spatial parallelism.


krober10nd:
Maybe we should also tackle receivers?

Olender:
> Maybe we should also tackle receivers?

I agree. Going to look into it as soon as all tests pass. Waiting for CircleCI

Olender:
Right now the new is_local inside receivers and sources has the same cell ids (just one per item) as the cell id in the tabulations. So even if a receiver is on an edge no error should occur

Olender:
Getting some weird "cannot allocate memory" in the test gradient pml

krober10nd:
Lets make sure the gradient tests for CI happen using P=1

spyro/receivers/Receivers.py
```diff
@@ -343,13 +340,13 @@ def __func_build_cell_tabulations_3D(self):
...
 
-    def apply_source_receivers(self, rhs_forcing, residual, IT, is_local):
+    def apply_source_receivers(self, rhs_forcing, residual, IT):
         """
         The adjoint operation of interpolation (injection)
         """
         for rid in range(self.num_receivers):
             value = residual[IT][rid]
-            if is_local[rid]:
+            if self.is_local[rid]:
```
krober10nd, line 349: Note that you'll have to call the `rhs_forcing.dat.data_with_halos` the same number of times (i.e., num_receivers times) in order to avoid deadlocks in parallel.

krober10nd:
In order to verify that both the source and receiver treatment are valid in parallel, run `test_gradient.py` and compare the FD estimated values and the gradient. They should be the same as serial otherwise something is wrong. The test `test_parallel_sources.py` just compares the solution against a serial simulation.

Olender:
> Lets make sure the gradient tests for CI happen using P=1 with a dt of 0.001 s

Done. All tests pass (in serial) now (at least locally)

krober10nd:
okay so in the forward we `interpolate` to a set of points using the definition of interpolation and we also use this to force the source term so should there be separate methods for `interpolate` and `apply_source`? Likewise for the adjoint, we call this `apply_something` perhaps this should be called `adjoint_interpolation`?

krober10nd:
also, ssprk came back here somehow will have to rebase with main before merging...

Olender:
> okay so in the forward we `interpolate` to a set of points using the definition of interpolation and we also use this to force the source term so should there be separate methods for `interpolate` and `apply_source`?

I think so. Interpolate gets a field as an input and interpolates those values into an arbitrary point (not necessarily a nodal point in the element). Apply_source doesn't take a field as a input, only a scalar value (or scalar value per element) and applies it, based on the cell tabulations to the nodal points.

> Likewise for the adjoint, we call this apply_something perhaps this should be called adjoint_interpolation?

I think this makes sense. The current name 'apply_source_receivers' is weird. However apply_receiver_as_source might make it more readable to an end user. I think both are good options

Olender:
> okay so in the forward we `interpolate` to a set of points using the definition of interpolation and we also use this to force the source term so should there be separate methods for `interpolate` and `apply_source`? Likewise for the adjoint, we call this `apply_something` perhaps this should be called `adjoint_interpolation`?

So apply_source is kind of the transverse of interpolate in receivers. It would make sense to rewrite the interpolation in order to have it also work as the adjoint_interpolation therefore removing one of those methods, but I think we can leave it for now.

krober10nd:
I like the more verbose name. But still have to figure out why we don’t get the same gradient in parallel tomorrow

krober10nd:
@Olender could you take a look at these two failing `new_at` tests?

Olender:
Yes, let me see what's happening

Olender:
> This is probably related to lowering the tolerance on the point locator method.

It's not the tolerance. Changing it to the previous value doesn't help. Going to continue debugging since it doesn't find the correct cell in 3D.

Olender:
It's a simple fix. The 2D locate cell was outside its if in build_maps

Olender:
It was a simple fix. Cell_id wasn't inside the 2d or 3d if clause, so it was always running as if in 2D.

PR #37, opened 2021-05-07
removing other damping layers

krober10nd:
* gets rid of unused damping layers and the unused formulation module for the old Newmark scheme (which is completely untested)


krober10nd:
@Olender

PR #38, opened 2021-05-07
New source options

Olender:
Adding option for applying differing sources per shot


krober10nd:
Could you write how you would call this for multiple sources in a fenced up block here?

krober10nd:
why was this closed?

Olender:
Made another branch and PR changing less stuff

PR #40, opened 2021-05-10
ensemble paralleism happens automatically

krober10nd:
* with the usage of decorators inside `forward` and `gradient`, shots are distributed across the ensemble communicator by default with the user doing anything.


krober10nd:
I'm adding similar capabilities for plotting.

PR #43, opened 2021-05-13
AD updates

Ig-dolci:


krober10nd:
could you rebase with main and I'll take another look at this? Thanks!

PR #48, opened 2021-06-07
Gradient verification with AD

Ig-dolci:
Adding test gradient with AD


Olender (commented):
Hi Daiane, would it be possible to add markdown cells with text in the .ipynb files to make full utilization of the Jupyter Notebook features?

Olender:
I am going to go ahead and delete the .egg-info files

Olender:
Honest question: why did you remove the empty lines from the documentation in the sources class? Is it a best practices kind of thing? Asking so I can change it in the future as well

Olender:
I would appreciate inputs from @krober10nd and Lucas on the forward_AD wrapping inside a new solvers_AD class

Ig-dolci:
I have a question for the Spyro developers, it is not related to this pull request. Why the tolerance was set at cell_id = mesh.locate_cell([source_z, source_x], tolerance=0.01) in the function source_dof_finder?

Olender:
Because when the tolerance was too low and a source was located very near to a cell edge .locate_cell wouldn't find it in any cells

Olender:
We used to leave a 1e-6 tolerance but that made not finding sources a frequent problem. Sometimes it found them sometimes not. I used to have my own .locate_cell method in sources, but it was slightly slower then the firedrake (or UFL) one. If it becomes a problem we can go back to it

Olender:
@Ig-dolci, do your branches not run the CircleCI tests automatically?

Olender:
@Ig-dolci can I test and merge?

spyro/receivers/Receivers.py
```diff
@@ -347,7 +333,21 @@ def __func_build_cell_tabulations_3D(self):
 
         return cell_tabulations
 
-
+    def setPointCloudRec(self, comm, paralel_z=True):
```
krober10nd, line 336: Can you follow the convention of snake case? https://www.google.com/search?client=safari&rls=en&q=snake+case&ie=UTF-8&oe=UTF-8

Ig-dolci: Ok. I'm modifying here.

Ig-dolci: done!

Olender:
I think we should go ahead and do the PR. It will be nice to have automatic adjoint in the main

Olender (approved):

PR #51, opened 2021-07-21
Tools overhaul

Olender:
Merging first iteration of tools package, which allows DoF density calculation, into main


krober10nd:
Let’s add SeismicMesh to the dependencies list for this package.

Olender (commented):
I installed SeismicMesh without dependencies to skip avoid mpi4py, which I had difficulties installing in CircleCI. Since we don't run any parallel tests in CircleCI I don't think this is a problem for now. I installed individually the other dependencies that SeismicMesh required which weren't installed previously in the docker (CGAL, meshio, pyamg and h5py).

PR #53, opened 2021-08-02
Read me update

Olender:
Adding new figure for scaling benchmark and the relevant .py files in the benchmark folder. However, I am still waiting for mintrop to validate the .py files in the benchmark folder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ Functionality
 Performance
 ===========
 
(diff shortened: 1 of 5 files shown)


Olender:
Since I pulled together different parts of my code for the paper for the benchmark .py files ideally I would run it with the same node configuration in mintrop to check if it had the same runtime results as the figure. However, since mintrop is crazy busy today I just have access to the intel-small node which, if I am not mistaken, has a different processors then the others. Therefore, I am only testing if it runs in parallel in mintrop and gives results in the same order of magnitude for the 15 core case

krober10nd:
That's okay, thanks! It looks great.

PR #65, opened 2022-02-02
Test ad 3 d

Ig-dolci:


Olender:
Sorry about the delay, I still have to finish setting up NDF's workstation to do the tests, instead of circleCI, in order to do 3D tests

PR #69, opened 2022-04-08
Create python-tests.yml

Olender:


Olender:
Added testing with a self hosted runner

PR #131, opened 2024-11-01
Implement Stacey boundary condition

SouzaEM:
This implementation includes:

- Stacey boundary condition itself
- Modification of field logger to also log functionals
- Evaluation of mechanical energies to evaluate the boundary condition
- Regression tests based on the mechanical energy
- Second order time discretization options for velocities (although they didn't improve performance)


spyro/io/field_logger.py
Olender, line 1: I really like what you did with the FieldLogger class. Since it has grown a bit and if you have time could you add documentation as well? I think later I will add, in another PR, a File class that writes to binary in cases were firedrake is installed without VTK.

Olender (approved):

spyro/io/field_logger.py
SouzaEM: So let's open another PR to address the binary format, the documentation, and other new features.

PR #134, opened 2024-11-07
Add Clayton-Engquist boundary condition for acoustic wave

SouzaEM:


spyro/io/model_parameters.py
Olender, line 1: Good idea having the default top be false

spyro/solvers/acoustic_solver_construction_no_pml.py
Olender, line 1: We need to ensure that if we use other meshing tools in the future, we always follow the numbering scheme used here (top=1, bottom=2, etc.). In the future, it may be helpful to add a check after loading the mesh to verify this consistency.

Olender (approved):

spyro/solvers/acoustic_solver_construction_no_pml.py
SouzaEM: As this verification is also necessary for elastic waves, I will merge this pull request and open a new issue for making sure the numbering scheme is appropriate for the mesh.

PR #140, opened 2025-01-23
Apptainer/Singularity containers

### LINKED ISSUE #138: Apptainer/Singularity container
This issue encompasses the following tasks:

- [x] Add instructions for creating Apptainer/Singularity container from the Docker images (converting the Docker image to Apptainer image is more productive because Apptainer does not support layers by design. Then, creating a definition file for Apptainer takes more time than creating Dockerfiles)
- [ ] Evaluate the possibility of using Apptainer image in CI
- [x] Add instructions for running programs in Mintrop

SouzaEM:
Closes #138


Olender:
Is our CI runner down again? I will see if I can get it running to test this

docker/Dockerfile
Olender, line 1: Are we incompatible with newer meshio?

apptainer/README.md
Olender, line 1: I specially like that there are instructions for using it in mintrop! Does it work well in parallel?

Olender (approved):

docker/Dockerfile
SouzaEM: I built a image without specifying a version of meshio and all tests run successfully. Actually, I don't remember anymore why I fixed meshio version. It may be related to other implementations I was doing. I will roll back this change.

apptainer/README.md
SouzaEM: Yes. I did not put the parallel test in the example because it fails when not running from spyro root directory. I still did not figure out why.

PR #145, opened 2025-02-26
Issue 0142 fixing failing pip install

Olender:
Fixes failing pip install and also makes spyro compatible with latest Firedrake API changes.


docker/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN . /home/firedrake/firedrake/bin/activate; pip3 install siphash24
 
 FROM spyro_base AS spyro_release
 
-RUN . /home/firedrake/firedrake/bin/activate; pip install git+https://github.com/Olender/spyro-1.git
+RUN . /home/firedrake/firedrake/bin/activate; pip install git+https://github.com/NDF-Poli-USP/spyro.git
```
connorjward, line 27: This will not work with the pip-installed Firedrake as we don't make a venv for you any more

Olender: Thanks! I will address this issue on a later PR, since this dockerfile uses firedrakeproject/firedrake:2024-09

PR #148, opened 2025-03-10
Preparing AD version for the paper

Ig-dolci:


Olender:
I thought I had updated the runner after TAO was added, but I'll install a fresh Firedrake and run the tests again. Are there any additional dependencies that need to be installed in the runner to ensure the TAOSolver call from PyAdjoint works correctly?

Ig-dolci:
> I thought I had updated the runner after TAO was added, but I'll install a fresh Firedrake and run the tests again. Are there any additional dependencies that need to be installed in the runner to ensure the TAOSolver call from PyAdjoint works correctly?

PETSc allows the use of TAO without any additional dependencies, which is great. Although I encountered issues using TAO for `L2` Riesz representation, I still believe that enabling TAO as an optimisation solver is attractive. Hopefully, the TAO developers will improve support for `L2` in the future 🤞🏼.

Olender:
Just to follow up. I still haven't finishing installing a newer Firedrake on our runner. IT at USP decided to do some maintenance on the workstations and I cannot access it by ssh anymore. I am still trying to regain access

PR #158, opened 2025-05-19
Adding installation instructions

Olender:
This PR adds installation instructions in the README file. It includes installation with or without optional dependencies. I've also removed our reference to the slack channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,21 @@ To implement these solvers, spyro uses the finite element package [Firedrake](ht
 
 To use spyro, you'll need to have some knowledge of Python and some basic concepts in inverse modeling relevant to active-source seismology.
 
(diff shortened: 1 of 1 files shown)


ruansava (changes requested):
@Olender Consider include netgen and slepc dependencies for habc
Additional dependencies about formats discussed previously are very specific. Then I believe we can look at other options within python and discard.

Olender:
Added SLEPC as a requirement on the instructions. The docker is also updated with the requirements (both SLEPC and NETGEN) in another branch. SHould I add netgen as a hard requirement or optional?

ruansava:
Hi @Olender 
I think that netgen should be a hard requirement at the moment. However, in the future, we can transform the code using netgen with utilities if meshio and seismicMesh (We can discuss this part with Thiago). 2D meshes work well with netgen but 3D meshes will definetively based on  meshio and seismicMesh

PR #162, opened 2025-05-26
161 move ci to docker

Olender:
Updates Docker to use latest Firedrake release and also moves our CI to it

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -6,14 +6,11 @@ RUN apt-get update \
     && apt-get -y install gmsh patchelf \
     && rm -rf /var/lib/apt/lists/*
 
-USER firedrake
 RUN pip3 install segyio
 
 # spyro dependencies
 USER root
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
Olender, line 1: Since the Firedrake release, and the spyro smoke tests in Firedrake, we can just use their Docker now. This docker also does not need a venv activation inside anymore

test/test_elastic_local_abc.py
Olender, line 1: @SouzaEM , can you tell me what AssertionError was raised before? I guess it was the energy instability one inside the forward_solve? I just can't reproduce this error anymore. Since last_mechanical_energy is less than expected I assume everything is working fine now.

SouzaEM: I didn't find in the code where the assertion was made. I think the problem was that discretizing the velocity with central difference could compromise the mass-lumping. The velocity appears in the local ABCs as far as I remember.

PR #167, opened 2025-06-18
Ruben/update habc

ruansava:
Improvements related to HABC Salas' method


spyro/habc/__init__.py
Olender, line 1: What does this do?

acoustic_energy.npy
Olender, line 1: Is this used for a test? If so we can move it to test/inputfiles

Olender:
I've gone ahead and removed the ipdb imports.

PR #193, opened 2025-11-14
Use native lump scheme

pbrubeck:


spyro/domains/quadrature.py
```diff
@@ -46,68 +44,11 @@ def quadrature_rules(V):
...
-            dimension=dimension, degree=degree
-        )
-        qr_s_rule = gauss_lobatto_legendre_cube_rule(
-            dimension=(dimension - 1), degree=degree
-        )
-        # Convert to dictionary format for consistent interface
-        qr_x = {"scheme": qr_x_rule}
-        qr_k = {"scheme": qr_x_rule}
-        qr_s = {"scheme": qr_s_rule}
+        qr_x = {"scheme": "lump", "degree": degree}
```
Ig-dolci: We could take advantage of this PR to document what is the range of degrees available for 2D and 3D quadratures.

PR #198, opened 2026-02-09
Rename from FE_element to function_space

Ig-dolci:
A more readable function that actually builds a function space.


Olender:
The slepc installation error is weird, I've tried running it again

PR #214, opened 2026-02-25
210 quicker ci

Olender:


.github/workflows/python-tests.yml
```diff
@@ -39,11 +40,11 @@ jobs:
         && apt-get -y install gmsh patchelf \
         && pip3 install segyio \
         && apt-get update && apt-get install -y libgmp3-dev libmpfr-dev libcgal-dev python3-tk \
-        && pip3 install --no-dependencies git+https://github.com/NDF-Poli-USP/SeismicMesh.git \
```
Olender: They are both the same thing and both work exactly the same

Tasinato:
Should the build workflow run only on Sundays, or should it also be triggered after the slow tests complete? 
Is there a specific condition that controls when the build should run?

PR #223, opened 2026-03-10
Dolci/auto adj

### LINKED ISSUE #300: mixed space assign|automated grads
Right update the wave solutions via assign operations for mixed space when using automated adjoints.

Ig-dolci:


spyro/solvers/acoustic_wave.py
```diff
@@ -91,14 +199,58 @@ def gradient_solve(self, guess=None, misfit=None, forward_solution=None):
...
-        else:
-            raise ValueError("Please load or calculate a real shot record first")
-        return backward_wave_propagator(self)
+        if guess is not None:
+            self.initial_velocity_model = guess
+
+        if compute_functional is not None:
+            self.compute_functional = compute_functional
+
+        observed_receivers = self._resolve_true_receivers(true_receivers)
```
Ig-dolci: Add a check for true_receivers data. Should be a list of numpy arrays or a list of firedrake function. The list length should be the total number of time steps.

Ig-dolci (commented):
ii

Ig-dolci (commented):
dd

spyro/solvers/acoustic_wave.py
```diff
@@ -77,28 +78,53 @@ def matrix_building(self):
...
+                automated_adjoint = AutomatedAdjoint(control)
+            self.automated_adjoint = automated_adjoint
+
+            previous_compute_functional = self.compute_functional
+            self.compute_functional = True
+            try:
+                with self.automated_adjoint.fresh_tape():
+                    self.automated_adjoint.start_recording()
+                    try:
+                        self.forward_solve(**kwargs)
```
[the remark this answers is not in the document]
Ig-dolci: For serial case, we should create a functional with two controls: c and source cofunction. This is because we can use the same tape to run the forward solver and update the source cofunction that is the unique difference

spyro/solvers/acoustic_wave.py
```diff
@@ -77,28 +78,53 @@ def matrix_building(self):
...
+                    finally:
+                        self.compute_functional = previous_compute_functional
+                        self.automated_adjoint.stop_recording()
+                    self.automated_adjoint.create_reduced_functional(
+                        self.functional
+                    )
+            finally:
+                self.compute_functional = previous_compute_functional
+            return self.automated_adjoint.compute_gradient()
```
Ig-dolci: We should compute the gradient only with respect to c

PR #259, opened 2026-04-02
fixing vom

Ig-dolci:
This PR seeks to fix VOM for the elastic solver, which I erroneously thought was working. I also added a test for the elastic and acoustic cases. The issue only appears in the Sources modelling.

### DIFF
--- a/spyro/sources/Sources.py
+++ b/spyro/sources/Sources.py
@@ -69,7 +69,7 @@ class Sources(Delta_projector):
 
         self.point_locations = wave_object.source_locations
         self.number_of_points = wave_object.number_of_sources
-        self.amplitude = self._normalize_amplitude(wave_object.amplitude)
+        self.amplitude = wave_object.amplitude
         self.is_local = [0] * self.number_of_points
         self.current_sources = None
         self.update_wavelet(wave_object)
@@ -77,12 +77,6 @@ class Sources(Delta_projector):
             self.build_maps(order=0)
         else:
             self.build_maps(order=1)
-
-    @staticmethod
-    def _normalize_amplitude(amplitude):
-        if np.isscalar(amplitude):
-            return amplitude
-        return np.asarray(amplitude)
 
     def update_wavelet(self, wave_object):
         self.wavelet = full_ricker_wavelet(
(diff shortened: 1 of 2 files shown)


spyro/sources/Sources.py
```diff
@@ -78,6 +78,12 @@ def __init__(self, wave_object):
         else:
             self.build_maps(order=1)
 
+    @staticmethod
+    def _normalize_amplitude(amplitude):
+        if np.isscalar(amplitude):
+            return amplitude
+        return np.asarray(amplitude)
+
```
Ig-dolci:
```suggestion

```

spyro/sources/Sources.py
```diff
@@ -122,22 +128,66 @@ def source_cofunction(self):
...
             self.function_space.dual()).interpolate(source_cofunction)
 
+    def _vertex_only_source_values(self):
+        num_sources = len(self.current_sources)
+
+        if self.wave_type == WaveType.ISOTROPIC_ACOUSTIC:
+            return np.full(num_sources, self._scalar_vertex_only_amplitude())
+
+        if self.wave_type == WaveType.ISOTROPIC_ELASTIC:
+            amplitude = self._elastic_vertex_only_amplitude()
```
Ig-dolci:
```suggestion
            amplitude = np.asarray(self.amplitude)
```

Ig-dolci (commented):
Avoid nested methods

tests/on_one_core/test_sources.py
```diff
@@ -58,5 +62,216 @@ def test_ricker_varies_in_time():
...
+    *,
+    dimension,
+    degree,
+    synthetic_data=None,
+    initial_velocity=None,
+):
+    receiver_location = (-0.2, 0.6) if dimension == 2 else (-0.2, 0.6, 0.6)
+    model = {
+        "options": {
+            "cell_type": "T",
```
Olender: It might be nice to add a fixture so these tests go through "T" and "Q" cells, what do you think?

Olender:
I approve everything in this PR. However, what would you think about adding one more test? Something like:

Ig-dolci:
> I approve everything in this PR. However, what would you think about adding one more test? Something like a fixture in the time_convergence test (tests/on_one_core/test_time_convergence.py):
> 
> ```
> import spyro
> import numpy as np
> import math
> import pytest
> 
> 
> def error_calc(p_numerical, p_analytical, nt):
>     norm = np.linalg.norm(p_numerical, 2) / np.sqrt(nt)
>     error_time = np.linalg.norm(p_analytical - p_numerical, 2) / np.sqrt(nt)
>     div_error_time = error_time / norm
>     return div_error_time
> 
> 
> @pytest.fixture(params=[False, True], ids=["no_vom", "vom"])
> def use_vertex_only_mesh(request):
>     return request.param
> 
> 
> def run_forward(dt, use_vertex_only_mesh=False):
>     # dt = float(sys.argv[1])
> 
>     final_time = 1.0
>     dx = 0.006546536707079771
> 
>     dictionary = {}
>     dictionary["options"] = {
>         "cell_type": "T",  # simplexes such as triangles or tetrahedra (T) or quadrilaterals (Q)
>         "variant": "lumped",  # lumped, equispaced or DG, default is lumped "method":"MLT", # (MLT/spectral_quadrilateral/DG_triangle/DG_quadrilateral) You can either specify a cell_type+variant or a method
>         "degree": 4,  # p order
>         "dimension": 2,  # dimension
>     }
> 
>     # Number of cores for the shot. For simplicity, we keep things serial.
>     # spyro however supports both spatial parallelism and "shot" parallelism.
>     dictionary["parallelism"] = {
>         "type": "automatic",  # options: automatic (same number of cores for evey processor) or spatial
>     }
> 
>     # Define the domain size without the PML. Here we'll assume a 1.00 x 1.00 km
>     # domain and reserve the remaining 250 m for the Perfectly Matched Layer (PML) to absorb
>     # outgoing waves on three sides (eg., -z, +-x sides) of the domain.
>     dictionary["mesh"] = {
>         "length_z": 3.0,  # depth in km - always positive
>         "length_x": 3.0,  # width in km - always positive
>         "length_y": 0.0,  # thickness in km - always positive
>         "mesh_file": None,
>         "mesh_type": "firedrake_mesh",  # options: firedrake_mesh or user_mesh
>     }
> 
>     # Create a source injection operator. Here we use a single source with a
>     # Ricker wavelet that has a peak frequency of 5 Hz injected at the center of the mesh.
>     # We also specify to record the solution at a microphone near the top of the domain.
>     # This transect of receivers is created with the helper function `create_transect`.
>     dictionary["acquisition"] = {
>         "source_type": "ricker",
>         "source_locations": [(-1.5 - dx, 1.5 + dx)],
>         "frequency": 5.0,
>         "delay": 0.3,
>         "receiver_locations": [(-1.5 - dx, 2.0 + dx)],
>         "delay_type": "time",
>     }
> 
>     if use_vertex_only_mesh:
>         dictionary["acquisition"]["use_vertex_only_mesh"] = True
> 
>     # Simulate for 2.0 seconds.
>     dictionary["time_axis"] = {
>         "initial_time": 0.0,  # Initial time for event
>         "final_time": final_time,  # Final time for event
>         "dt": dt,  # timestep size
>         "amplitude": 1,  # the Ricker has an amplitude of 1.
>         "output_frequency": 100,  # how frequently to output solution to pvds
>         "gradient_sampling_frequency": 100,  # how frequently to save solution to RAM
>     }
> 
>     dictionary["visualization"] = {
>         "forward_output": True,
>         "forward_output_filename": "results/forward_output.pvd",
>         "fwi_velocity_model_output": False,
>         "velocity_model_filename": None,
>         "gradient_output": False,
>         "gradient_filename": None,
>     }
> 
>     Wave_obj = spyro.AcousticWave(dictionary=dictionary)
>     Wave_obj.set_mesh(input_mesh_parameters={"edge_length": 0.02, "periodic": True})
> 
>     Wave_obj.set_initial_velocity_model(constant=1.5)
>     Wave_obj.forward_solve()
> 
>     rec_out = Wave_obj.receivers_output
> 
>     return rec_out
> 
> 
> @pytest.mark.slow
> def test_second_order_time_convergence(use_vertex_only_mesh):
>     """Test that the second order time convergence
>     of the central difference method is achieved"""
> 
>     dts = [
>         0.0005,
>         0.0001,
>     ]
> 
>     analytical_files = [
>         "tests/inputfiles/analytical_solution_dt_0.0005.npy",
>         "tests/inputfiles/analytical_solution_dt_0.0001.npy",
>     ]
> 
>     numerical_results = []
>     errors = []
> 
>     for i in range(len(dts)):
>         dt = dts[i]
>         rec_out = run_forward(dt, use_vertex_only_mesh=use_vertex_only_mesh)
>         rec_anal = np.load(analytical_files[i])
>         time = np.linspace(0.0, 1.0, int(1.0 / dts[i]) + 1)
>         nt = len(time)
>         numerical_results.append(rec_out.flatten())
>         errors.append(error_calc(rec_out.flatten(), rec_anal, nt))
> 
>     theory = [t**2 for t in dts]
>     theory = [errors[0] * th / theory[0] for th in theory]
> 
>     assert math.isclose(np.log(theory[-1]), np.log(errors[-1]), rel_tol=1e-2)
> 
> 
> if __name__ == "__main__":
>     test_second_order_time_convergence(False)
> ```

Acho que podemos adicionar esse teste em outro PR pra testar receptores.

> We need to update the DiracDelta class to just wrap over your VOM implementation in the future and eventually just get rid of the old source/receiver class all together

Joia. Vou criar uma issue pra isso.

tests/on_one_core/test_sources.py
```diff
@@ -58,5 +62,216 @@ def test_ricker_varies_in_time():
...
+        },
+        "mesh": _base_mesh(dimension),
+        "acquisition": {
+            "source_type": "ricker",
+            "source_locations": source_locations,
+            "frequency": 5.0,
+            "delay": 1.0,
+            "amplitude": amplitude,
+            "receiver_locations": [receiver_location],
+            "use_vertex_only_mesh": True,
```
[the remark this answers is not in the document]
Ig-dolci: Podemos criar uma issue pra tornar VOM default.

Olender (approved):

PR #260, opened 2026-04-06
Olender/fix mesh set and get

Olender:
This PR accomplishes 3 things:
- Cleans up the MeshingParameters class so it's easier to follow for integration with both Romildo's work and Ruben's HABC meshing methods;
-  Updates the automatic meshing function to skip and print if the meshing parameters class is incomplete;
- Calls the mesher during the get_mesh function call.


ruansava (approved):

ruansava:
Can we delete it?

PR #270, opened 2026-04-15
DO NOT SQUASH: Added Gmsh 2d Meshing

rasjr1305:
Added Gmsh 2d Meshing for Seismic problems. Some features include:

Added corrections for files with less than 1000 lines;
Added parameter for uniform structured mesh or winslow smoothed;
Delimitation of water interface, being detected with the velocity model;
User can generate the mesh with or without water interface;
User can choose between rectangular and elliptical padding geometry;
User can choose between unstructured triangular or structured quad winslow smoothed meshes;
For unstructured meshes, the velocity model can be extended to the padding or a constant value can be chosen;
Mesh is exported in msh format;

For testing please use this python script:

The marmousi model used can be found here 
https://s3.amazonaws.com/open.source.geoscience/open_data/elastic-marmousi/elastic-marmousi-model.tar.gz

```
import sys
import os
import matplotlib.pyplot as plt
import firedrake as fire
from spyro.meshing.meshing_parameters import MeshingParameters
from spyro.meshing.meshing_functions import AutomaticMesh

marmousi_params = {
    "mesh_type": "gmsh_mesh", # Type of automatic mesh
    "dimension": 2, # Dimension of the mesh
    
    # Dimensions 
    "length_z": 3500.0,   
    "length_x": 17000.0,
    "output_filename": "marmousi.msh",
    
    # SEGY / Seismic constraints
    "velocity_model": "MODEL_S-WAVE_VELOCITY_1.25m.segy", # Velocity model segy file
    "cells_per_wavelength": 2.0, 
    "source_frequency": 3.0,
    
    # Padding Parameters
    "padding_type": "rectangular" , # Padding types "rectangular" "hyperelliptical" None 
    "padding_x": 1500.0, # Padding size in x direction
    "padding_z": 1500.0, # Padding size in z direction
    "hyper_n": 3.0, # Hyperellipse exponent
    "hmin_segy": 0.0, # Minimum Element size for segy, will apply if higher than function minimum
    "grade": 0.9, # function grading for smooth element transition, None = no smooth, 0.9 = small smooth, 0.1 = high smooth
    
    # Water Interface
    "water_interface": True, # If True detect and implement water interface
    "water_search_value": 0.0, # If enabled water interface, search for this water value to make the interface
    "vp_water": 500.0, # Substitute Water speed for this value if vs = 0.0
    
    # Structured Mesh & Winslow Smoothing
    "structured_mesh": True, # True if structured quad mesh, False if triangular unstructured mesh
    "min_element_size": 50.0, # Element size for structured mesh
    "apply_winslow": True, # If True apply winslow smoothing
    "winslow_implementation": "fast", # Winslow version to use, default, fast and numba are options
    "winslow_iterations": 300, # Number of iterations for Winslow Smoothing
    "winslow_omega": 0.5, # Winslow Smoothing node movement factor
    "extend_segy": False, # Extend the segy function into the padding ( for unstructured mesh )
    "h_padding": 500.0 # If extend_segy = False, use this value of constant padding size
}

mesh_params = MeshingParameters(input_mesh_dictionary=marmousi_params, velocity_model=marmousi_params["velocity_model"])

mesh_generator = AutomaticMesh(mesh_parameters=mesh_params)

print("Starting Gmsh mesh generation...")
firedrake_mesh = mesh_generator.create_mesh()

if firedrake_mesh is not None:
    print(f"Mesh successfully generated and loaded from {mesh_params.output_filename}!")
    
    fig, axes = plt.subplots(figsize=(60, 60))
    fire.triplot(firedrake_mesh, axes=axes)
    axes.set_aspect('equal')
    axes.set_title("Marmousi Mesh")
    axes.set_xlabel("Distance X (m)")
    axes.set_ylabel("Depth Z (m)")
    plt.show()
```


spyro/meshing/meshing_functions.py
```diff
@@ -16,6 +16,18 @@
 except ImportError:
     SeismicMesh = None
 
+try:
```
Olender, line 19: Why have you added a try except block here? The SeimicMesh and gmsh ones on top were only added because SeismicMesh and gmsh are optional dependencies. I think some of these methods here have gmsh and/or numba dependencies and some don't, but you seem to have already added the optional import there. 

I still have to revise your other files, but I do prefer a single point of failure with a clear error or message if possible

spyro/meshing/meshing_functions.py
```diff
@@ -485,6 +501,950 @@ def create_spyro_mesh(self):
...
+        """
+        Creates a 2D mesh using Gmsh with optional water interface, 
+        elliptical/rectangular padding, and Winslow smoothing.
+
+        Returns
+        -------
+        mesh : Firedrake Mesh
+            The loaded Firedrake mesh object.
+        """
+        if gmsh is None:
```
Olender, line 514: Perfect

spyro/meshing/meshing_functions.py
```diff
@@ -485,6 +501,950 @@ def create_spyro_mesh(self):
...
+            raise ImportError("gmsh is not available. Please install it.")
+
+        # MPI Setup: Only generate the mesh on rank 0
+        rank = 0
```
Olender, line 518: We have a decorator now that does this, so devs don't have to deal with comm in comm = 0 required methods

spyro/meshing/meshing_functions.py
```diff
@@ -485,6 +501,950 @@ def create_spyro_mesh(self):
...
+        if self.comm is not None:
+            if hasattr(self.comm, 'ensemble_comm'):
+                rank = self.comm.ensemble_comm.rank
+            else:
+                rank = self.comm.comm.rank
+
+        if rank == 0:
+            parallel_print("Generating Gmsh mesh on Rank 0...", comm=self.comm)
```
Olender, line 526:
```suggestion
        parallel_print("Generating Gmsh mesh on Rank 0...", comm=self.comm)
```
parralel_print deals with comm internally

spyro/meshing/meshing_parameters.py
```diff
@@ -164,8 +196,8 @@ def __init__(self, input_mesh_dictionary=None, dimension=None, source_frequency=
         degree : int, optional
             Polynomial degree of finite element basis functions.
             Default is None.
-        velocity_model : object, optional
-            Velocity model object for mesh adaptation. Default is None.
+        velocity_model : object or str, optional
```
Olender, line 199: We do handling of the velocity model input (if its segy, bin, hdf5, firedrake function, etc.) in another part of the code

spyro/meshing/meshing_parameters.py
```diff
@@ -194,19 +226,39 @@ def __init__(self, input_mesh_dictionary=None, dimension=None, source_frequency=
...
 
         # Set basic attributes
         self.quadrilateral = quadrilateral
         self.method = method
         self.minimum_velocity = None
-        self.velocity_model = velocity_model
         self.gradient_mask = None
         self.negative_z = negative_z
 
+        self.velocity_model = self.input_mesh_dictionary.get("velocity_model", velocity_model)
```
Olender, line 244: I prefer we go back to self.velocity_model = velocity_model since this input is tricky and dealt with elsewhere

spyro/meshing/meshing_parameters.py
```diff
@@ -194,19 +226,39 @@ def __init__(self, input_mesh_dictionary=None, dimension=None, source_frequency=
...
+        self.minimum_velocity = self.input_mesh_dictionary.get("minimum_velocity", None)
+        self.ellipse_n = self.input_mesh_dictionary.get("ellipse_n", 3.0)
+        self.hmin_segy = self.input_mesh_dictionary.get("hmin_segy", 0.0)
+        self.grade = self.input_mesh_dictionary.get("grade", 0.9)
```
Olender, line 248: Perfect! But why did you choose 0.9? I never know what to pick here

spyro/meshing/meshing_parameters.py
```diff
@@ -125,14 +125,46 @@ class MeshingParameters():
...
+    ellipse_n : float
+        Ellipse exponent for elliptical padding.
+    hmin_segy : float
+        Minimum Element size for SEGY interpolation.
+    grade : float
+        Function grading for smooth element transition.
+    water_interface : bool
+        If True, detects and implements the water interface curve.
+    water_search_value : float
+        The physical value to search for to trace the water interface.
```
Olender, line 144: Could you detail this a bit better? I couldn't understand exactly what this attribute is

spyro/meshing/meshing_parameters.py
```diff
@@ -125,14 +125,46 @@ class MeshingParameters():
...
+    h_padding : float
+        Constant padding element size used if extend_segy is False.
 
     Notes
     -----
     The class enforces mutual exclusivity between 'edge_length' and
     'cells_per_wavelength' parameters. Setting one will clear the other.
 
-    Mesh dimensions (length_x, length_y, length_z, abc_pad_length) are
+    Mesh dimensions (length_x, length_y, length_z, abc_pad_length, padding_x, padding_z) are
```
Olender, line 167: Is it okay with you if I change the setters and getters of padding_x and padding_z to set them equal to abc_pad_length if no value is given to them?

spyro/meshing/meshing_parameters.py
```diff
@@ -194,19 +226,39 @@ def __init__(self, input_mesh_dictionary=None, dimension=None, source_frequency=
...
+        self.hmin_segy = self.input_mesh_dictionary.get("hmin_segy", 0.0)
+        self.grade = self.input_mesh_dictionary.get("grade", 0.9)
+        self.water_interface = self.input_mesh_dictionary.get("water_interface", False)
+        self.water_search_value = self.input_mesh_dictionary.get("water_search_value", 0.0)
+        self.vp_water = self.input_mesh_dictionary.get("vp_water", 1500.0)
+        self.structured_mesh = self.input_mesh_dictionary.get("structured_mesh", False)
+        self.min_element_size = self.input_mesh_dictionary.get("min_element_size", 35.0)
+        self.winslow_iterations = self.input_mesh_dictionary.get("winslow_iterations", 5000)
+        self.winslow_omega = self.input_mesh_dictionary.get("winslow_omega", 0.5)
+        self.extend_segy = self.input_mesh_dictionary.get("extend_segy", True)
```
Olender, line 256: This is for velocity model extension right?

spyro/meshing/meshing_parameters.py
```diff
@@ -825,6 +905,150 @@ def periodic(self, value):
...
+
+        Returns
+        -------
+        str or None
+            The implementation method ('default', 'fast', 'numba', or None).
+        """
+        return self._winslow_implementation
+
+    @winslow_implementation.setter
+    def winslow_implementation(self, value):
```
Olender, line 1034: Yes! I am not the only one in this project that likes to use setters and getters any more!

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
+import numpy as np
+import gmsh
+from scipy.signal import savgol_filter
+from scipy.interpolate import RegularGridInterpolator
```
Olender, line 4:
```suggestion
import numpy as np
import gmsh
import segyio
from scipy.signal import savgol_filter
from scipy.interpolate import RegularGridInterpolator
from pathlib import Path
from typing import Callable, Tuple
from scipy.interpolate import griddata

try:
    from numba import njit, prange
    NUMBA_AVAILABLE = True
except ImportError:
    NUMBA_AVAILABLE = False
```

Lets put all the imports in the top

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
...
+    rows, cols = grid_values.shape
+    print(window_length_z,window_length_x)
+    # First filter along axis 0 (columns), then along axis 1 (rows)
+    filtered_values = savgol_filter(grid_values, window_length_x, polyorder, axis=0)
+    filtered_values = savgol_filter(filtered_values, window_length_z, polyorder, axis=1)
+    
+    return filtered_values
+
+from pathlib import Path
+import segyio
```
Olender, line 263:
```suggestion
```

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
...
+
+    # Calculate Final Shifted Coordinates 
+    x_shift = float(min(x_min, x_max))
+    Xs = Xs + x_shift  
+
+    return Xs, Z_bottom
+
+
+from typing import Callable
+from scipy.interpolate import griddata
```
Olender, line 389:
```suggestion
```

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
...
+                    
+                if is_all or is_Z:
+                    Z_new[i] = (1.0 - omega) * Z[i] + omega * Z_target
+
+        X[:] = X_new[:]
+        Z[:] = Z_new[:]
+
+    return np.column_stack((X, Z))
+
+from typing import Callable
```
Olender, line 622:
```suggestion
```

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
...
+
+# Attempt to import Numba
+try:
+    from numba import njit, prange
+    NUMBA_AVAILABLE = True
+except ImportError:
+    NUMBA_AVAILABLE = False
+
```
Olender, line 630:
```suggestion
```

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
...
+        )   
+    def bilinear_interp_2d(*args, **kwargs):
+        raise ImportError(
+            "Numba is not installed in this environment. "
+            "Please run 'pip install numba' to use the 'numba' implementation, "
+            "or switch configuration to winslow_implementation='fast' or 'default'."
+        )
+
+import numpy as np
+from typing import Callable, Tuple
```
Olender, line 1113:
```suggestion
```

spyro/meshing/meshing_parameters.py
rasjr1305: Yes !

spyro/meshing/meshing_parameters.py
rasjr1305: This is in case the user wants to extend the velocity model to the padding or if the user wants a constant element size in the padding

spyro/meshing/meshing_functions.py
ruansava: I agree

spyro/meshing/meshing_functions.py
```diff
@@ -485,6 +501,950 @@ def create_spyro_mesh(self):
         else:
             raise ValueError("dimension is not supported")
 
+    def create_gmsh_2D_mesh(self):
+        """
+        Creates a 2D mesh using Gmsh with optional water interface, 
+        elliptical/rectangular padding, and Winslow smoothing.
```
ruansava, line 507: Replace elliptical with hyperelliptical where applicable

spyro/meshing/meshing_functions.py
```diff
@@ -485,6 +501,950 @@ def create_spyro_mesh(self):
...
+                
+                    gmsh.model.occ.synchronize()
+                    gmsh.model.occ.removeAllDuplicates()
+                    gmsh.model.occ.synchronize()
+                
+                    gmsh.model.addPhysicalGroup(2, [water_surface], name="WaterSurface")
+                    gmsh.model.addPhysicalGroup(2, [surf_rock], name="SubSurface")
+                    gmsh.model.addPhysicalGroup(2, [surf_pad_tr, surf_pad_mr, surf_pad_tl, surf_pad_ml, surf_pad_bl, surf_pad_bc, surf_pad_br], name="Padding")
+                
+                if padding_type == "elliptical":
```
ruansava, line 676: "hyperelliptical"

spyro/meshing/meshing_functions.py
```diff
@@ -485,6 +501,950 @@ def create_spyro_mesh(self):
...
+                    a_val = (length_x / 2.0) + padding_x
+                    b_val = abs(depth_z / 2.0) + padding_z
+                    
+                    def intersect(x0, z0, dx, dz):
+                        def f(s):
+                            x, z = x0 + s * dx, z0 + s * dz
+                            return (abs(x - xc) / a_val)**ellipse_n + (abs(z - zc) / b_val)**ellipse_n - 1.0
```
ruansava, line 683: Replace ellipse_n  with hyper_n

spyro/meshing/meshing_parameters.py
ruansava: I haven't learned to do it, but I hope to include it in my programming skills

spyro/meshing/meshing_functions.py
ruansava, line 1: This file is too long. I think following a standard of around 1.000 lines per file is a good practice.

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,1719 @@
...
+        omega : float, optional
+            Relaxation factor for node movement (0 < omega <= 1). Default is 0.05.
+
+        Returns
+        -------
+        ndarray
+            Updated node coordinates array of shape (N, 2).
+        """
+        
+        if move_ellipse is None:
```
ruansava, line 977: Replace hyperell_params with ellipse_params where applicable

spyro/meshing/meshing_utils.py
ruansava, line 1: This file is too long. I think following a standard of around 1.000 lines per file is a good practice.

ruansava (changes requested):

spyro/meshing/meshing_utils.py
Olender: I think we can just send some of these methods 'as is' to other places. read_segy_velocity_model should be in io, either in basicio or material_properties_io

Olender: However, `calculate_edge_length` should be moved from `meshing_functions` to here. Since it's basicaly the oposite of `calculate_wavelength_sizing`

Olender: We can also move `vectorized_bilinear_interp`to general utils, since it can be usefull eslewhere. @rasjr1305, what do you think?

rasjr1305: The function vectorized_bilinear_interp and the other bilinear_interp are used to make a fast interpolation of the sizing function ( calling the direct python function can be very slow, you can test it using the default winslow smoothing method ), so wherever we need to obtain sizing values faster than calling the direct sizing function, the vectorized_bilinear_interp  can be used.

rasjr1305: > I think we can just send some of these methods 'as is' to other places. read_segy_velocity_model should be in io, either in basicio or material_properties_io

Yes some methods in the meshing utils are for reading, interpolating and grading the 2d sizing function, maybe we could place them elsewhere ?

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,612 @@
...
+        A tuple containing:
+        - sizing_function (callable): A function f(x, y) returning element size.
+        - min_val (float): Minimum element size in the domain.
+        - max_val (float): Maximum element size in the domain.
+        - n_samples (int): Number of depth samples.
+        - n_traces (int): Number of lateral traces.
+    """
+
+    # Read velocity model with provided bbox
+    vp, n_samples, n_traces = read_segy_velocity_model(fname)
```
Olender, line 170: Let's let this in, but we should put segy io outside of here

Olender: Let's create an issue

Olender:
I am trying to figure out what is happening in both the vp io and the wave propagation in olender_romildo/mesh_tester

Olender:
Is the first axis in the mesh z? If not can you rotate it? When aprojecting the velocity model from the file into the mesh I am getting 0 values everywhere

spyro/meshing/meshing_parameters.py
```diff
@@ -195,18 +229,40 @@ def __init__(self, input_mesh_dictionary=None, dimension=None, source_frequency=
...
         self.negative_z = negative_z
+        self.velocity_model = velocity_model
+
+        self.minimum_velocity = self.input_mesh_dictionary.get("minimum_velocity", None)
+        self.hyper_n = self.input_mesh_dictionary.get("hyper_n", 3.0)
+        self.hmin_segy = self.input_mesh_dictionary.get("hmin_segy", 0.0)
+        self.grade = self.input_mesh_dictionary.get("grade", 0.9)
+        self.water_interface = self.input_mesh_dictionary.get("water_interface", False)
+        self.water_search_value = self.input_mesh_dictionary.get("water_search_value", 0.0)
+        self.vp_water = self.input_mesh_dictionary.get("vp_water", 1500.0)
```
Olender: Makes sense to add this, but you've defined it again in the mesher with a default value there as well. I will point it out, but we should define it once here

spyro/io/material_properties_io.py
```diff
@@ -554,3 +554,15 @@ def set_tensor_function_space(wave, shape_func_space, is_dg):
             shape=shape_func_space,
         )
         return V
+
+
+class VelocityModelFileIO:
```
ruansava, line 559: The class should be renamed to assign general scalar properties, and try the velocity model as a case where the class can be used

ruansava: Another idea is that it is a wrapper of the general class for I/O properties

spyro/meshing/meshing_functions.py
ruansava, line 51: Is it possible to use some methods decoupled from the class? If so, they should be moved to the tools folder in a file called tools_gmsh and then imported here.

rasjr1305: Its a good idea if these methods could be moved for their own files, ex. firedrake_mesh to his own file, gmsh_mesh for its own file and then imported there, however this would be big refactor. Maybe we make the PR work as intended then flag for this refactor what do you think ?

ruansava: I agree. Maybe in the next PR

spyro/meshing/meshing_functions.py
```diff
@@ -551,6 +556,173 @@ def create_spyro_mesh(self):
...
+        if self.comm is None or self.comm.ensemble_comm.rank == 0:
+            parallel_print("Generating Gmsh mesh...", comm=self.comm)
+
+            depth_z = -abs(self.length_z)
+            length_x = self.length_x
+            padding_z = self.mesh_parameters.padding_z
+            padding_x = self.mesh_parameters.padding_x
+            hyper_n = self.mesh_parameters.hyper_n
+            fname = self.velocity_model
+            hmin_segy = self.mesh_parameters.hmin_segy
```
Olender, line 499: Where is hmin_segy set? And why isn't it just hmin?

spyro/meshing/meshing_functions.py
```diff
@@ -149,6 +149,7 @@ def create_mesh(self):
         mesh : Mesh
             Mesh
         """
+        self.mesh_parameters.check_completeness()
```
ruansava, line 161: Does "if self.mesh_parameters.check_completeness():" work?

spyro/io/basicio.py
```diff
@@ -346,12 +346,15 @@ def write_function_to_grid(function, V, grid_spacing, buffer=False):
...
+    num_grid_x = int(round((max_x - min_x ) / grid_spacing, 0)) + 1
+    num_grid_y = int(round((max_y - min_y ) / grid_spacing, 0)) + 1
+    xi = np.linspace(min_x, max_x, num_grid_x)
+    yi = np.linspace(min_y, max_y, num_grid_y)
     if dimension == 2:
         xi, yi = np.meshgrid(xi, yi)
     elif dimension == 3:
-        zi = np.arange(min_z, max_z, grid_spacing)
+        num_grid_z = int(round((max_z - min_z ) / grid_spacing, 0)) + 1
+        zi = np.arange(min_z, max_z, num_grid_z)
```
ruansava: Why are xi and yi calculated with linespace but zi with arange?

testing_mesher_with_velocity_grid.py
ruansava, line 1: Could this be an example?

ruansava (approved):
The changes to be implemented would be minimal, so that I can approve this

spyro/io/material_properties_io.py
ruansava: Could you please create an issue for this?

rasjr1305:
> Is the first axis in the mesh z? If not can you rotate it? When aprojecting the velocity model from the file into the mesh I am getting 0 values everywhere

The 2d mesh is now rotated to work with Spyro defined axis, maybe later we can add an "axis" option and the mesh is rotated according to user setting

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,612 @@
...
+            Coordinates where to evaluate element size
+
+        Returns:
+        --------
+        float or array
+            Element size at the given position(s)
+        """
+        return interpolate_size(x, y, cell_size, bbox)
+
+    return sizing_function, cell_size.min(), cell_size.max(), n_samples, n_traces
```
brunomariz, line 288: I believe returning a function in this case is unnecessary, since `sizing_function` is one line long (`return interpolate_size(x, y, cell_size, bbox)`). Perhaps a better practice would be to return the values used in the inner `sizing_function` function, and call `interpolate_size` directly where you need it.

Also, I don't understand why this function returns `cell_size.min()` and `cell_size.max()`, if they are obtainable from `calculate_wavelength_sizing(vp, wl, freq)`. @Olender observed that `cell_size` is being altered after it's creation in this function (e.g. `cell_size = np.pad(cell_size, padding, "edge")`, `cell_size = apply_savitzky_golay_filter_2d(cell_size, window_length_x, window_length_z)`), but these alterations could be made from a single function, such as `calculate_wavelength_sizing`.

Overall, I have the impression that this function is responsible for too many tasks, and I'd suggest trying to only have a single goal per function, since this makes code easier to read and debug.

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,612 @@
...
+
+
+def read_segy_velocity_model(fname):
```
brunomariz, line 291: It was already pointed out that this function could perhaps be in some other IO-related module.

I also wanted to ask whether it would be possible to make this a bit more general. It's not immediately clear to me why this would only work for velocity models. Could this function be generalized into something like the code below? Could this be useful for other models besides velocity?

```python3
def read_segy_model(fname):
    with segyio.open(fname, 'r', ignore_geometry=True) as segy:
        n_traces = len(segy.trace)
        n_samples = len(segy.samples)
        # Read traces directly into array
        model = np.zeros((n_samples, n_traces))
        for i in range(n_traces):
            model[:, i] = segy.trace[i]
    print(f"Final model range: {model.min():.1f} - {model.max():.1f}")
    return model, n_traces, n_samples
```

testing_mesher_with_velocity_grid.py
Olender: We are working on it

spyro/meshing/meshing_functions.py
Olender: Yes

spyro/io/basicio.py
Olender: Because I haven't changed it yet

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,612 @@
...
+try:
+    import gmsh
+except ImportError:
+    gmsh = None
+
+
+def check_gmsh(func):
+    """Decorator for gmsh check.
+
+    If gmsh isn't available returns ImportError
```
brunomariz: This comment may be too petty but it actually doesn't return `ImportError`, it raises an `ImportError`

spyro/meshing/meshing_parameters.py
Olender: We should centralize pad extension in another issue. Since both you and @ruansava made methods that do the same thing

Olender (changes requested):

testing_mesher_with_velocity_grid.py
Olender: I am currently trying to test Romildo's meshe generation with a spyro generated grid (instead of a segy file). Afterwards this will turn into a test and/or an example

spyro/io/basicio.py
Olender: Thanks for pointing it out. I've changed it

spyro/meshing/meshing_utils.py
```diff
@@ -0,0 +1,612 @@
...
+            is_left = np.all(np.abs(xs - 0.0) < tol)
+            is_right = np.all(np.abs(xs - length_x) < tol)
+            if not (is_top or is_left or is_right):
+                for t in n_tags:
+                    if t in tag_to_index:
+                        interface_nodes.add(tag_to_index[t])
+    return interface_nodes
+
+
+def align_water_columns_to_interface_x(points_2d, water_surface_nodes, interface_nodes, quads):
```
brunomariz, line 748: This function seems a bit long, could be broken down into other smaller functions? I do not understand the algorithms used here, so I cannot provide more specific suggestions, just this observation.

brunomariz (dismissed):
Submitting this review after adding comments to `meshing_utils.py`

spyro/meshing/meshing_utils.py
brunomariz: In fact, the only place where `create_sizing_function` is called, in the file `meshing_functions.py` is ignoring the returned values of `cell_size.min()` and `cell_size.max()`. This is a strong indicator that this function is returning more objects than necessary. Here's the code that uses `create_sizing_function`. In the rest of the code, `f_min` and `f_max` are never used.

```python
            ef_segy2, f_min, f_max, n_samples, n_traces = create_sizing_function(
                fname=fname, hmin=hmin_segy, bbox=segy_bbox, wl=wl, freq=freq,
                pad_type=padding_type, pad_size_x=padding_x, pad_size_z=padding_z,
                grade=grade, vp_water=vp_water
            )
```

spyro/meshing/meshing_functions.py
rasjr1305: This parameter is to set a minimum element size in the interpolation function, it will only apply if its higher than the function minimum

spyro/meshing/meshing_utils.py
rasjr1305: You are right, this function will work the same if we return only the interpolator, I will make the adjustments

spyro/meshing/meshing_utils.py
rasjr1305: We could create a separate framework only for IO and interpolation, I think we could create an Issue and solve this after this PR is approved ?

spyro/meshing/meshing_utils.py
rasjr1305: The return of minimum and maximum values yes could be removed, the number of samples and traces are used in later

spyro/meshing/meshing_utils.py
rasjr1305: Corrected !

spyro/meshing/meshing_utils.py
rasjr1305: In Gmsh, when we use the transfinite algorithm, if the structured lines go from a curve to a straight line, they tend to not be perfectly vertical aligned. This can happen when the water interface is very curved, so this function takes the inclined lines and make then vertical. I can take a look and see where this function can be improved

Olender:
@rasjr1305 , I've added two files in the root directory (that we should remake into tests before merging in). Please have a look at `propagating_mesh_with_avenir_segy.py` and `propagating_mesh_with_conditional_vp.py`

spyro/meshing/meshing_gmsh2d.py
```diff
@@ -0,0 +1,764 @@
...
+    parameters needed for subsequent mesh smoothing operations.
+    """
+    xc = length_x / 2.0
+    zc = depth_z / 2.0
+    a_val, b_val = None, None
+
+    pad_x_min, pad_x_max, pad_z_min = None, None, None
+    z_water_L, z_water_R = None, None
+
+    if padding_type == "hyperelliptical":
```
brunomariz, line 25: I believe `..._type` parameters add confusion to code, especially when these parameters are not type hinted, since the programmer has to read the function in order to know which "... types" (padding types, in this case) are allowed. 

I would suggest that instead of using a `padding_type` parameter, one function is created for each type of padding. For example, one function could be named `build_gmsh_geometry_and_groups_hyperellyptical`, and another function could be named `build_gmsh_geometry_and_groups_rectangular`.

The parts that are the same for all padding types can be each their own function too.

spyro/meshing/meshing_gmsh2d.py
```diff
@@ -0,0 +1,764 @@
...
+
+            gmsh.model.occ.synchronize()
+            gmsh.model.occ.removeAllDuplicates()
+            gmsh.model.occ.synchronize()
+
+            gmsh.model.addPhysicalGroup(2, [surf_internal], name="SubSurface")
+            gmsh.model.addPhysicalGroup(2, [surf_pad_left, surf_pad_bl, surf_pad_bot, surf_pad_br, surf_pad_right], name="Padding")
+
+        if padding_type == "hyperelliptical":
+            def intersect(x0, z0, dx, dz):
```
brunomariz: Is there a reason why this is being defined as an inner function?

Perhaps it would be best to define an `intersect` function elsewhere, perhaps some utils folder. That way, it could be used in other places too, if it is useful.

I also try to avoid defining inner functions since it is harder to write tests for them.

spyro/meshing/meshing_gmsh2d.py
```diff
@@ -0,0 +1,764 @@
...
+
+            ray_TL = gmsh.model.occ.addLine(pt_tl, pt_O_TL)
+            ray_BL = gmsh.model.occ.addLine(pt_bl, pt_O_BL)
+            ray_BML = gmsh.model.occ.addLine(pt_bot_midL, pt_O_BML)
+            ray_BMR = gmsh.model.occ.addLine(pt_bot_midR, pt_O_BMR)
+            ray_BR = gmsh.model.occ.addLine(pt_br, pt_O_BR)
+            ray_TR = gmsh.model.occ.addLine(pt_tr, pt_O_TR)
+
+            def make_arc(p1_tag, p2_tag, x1, z1, x2, z2, num_pts=25):
+                def get_theta(x, z):
```
brunomariz: Here there's an inner function inside of an inner function. The same comment as before applies, but in this case this level of nesting results in less readable code. In general, it is good to avoid nesting as much as possible for readability. That includes nested loops and conditional statements as well.

spyro/meshing/meshing_gmsh2d.py
```diff
@@ -0,0 +1,764 @@
...
+                gmsh.model.mesh.setTransfiniteSurface(surf_pad_right, cornerTags=[pt_tr, pt_br, pt_O_BR, pt_O_TR])
+
+                padding_surfs = [surf_pad_left, surf_pad_bot_left, surf_pad_bot_mid, surf_pad_bot_right, surf_pad_right]
+                for surf in padding_surfs:
+                    gmsh.model.mesh.setRecombine(2, surf)
+
+                gmsh.model.mesh.setTransfiniteSurface(surf_rock, cornerTags=[pt_tl, pt_tr, pt_br, pt_bl])
+                gmsh.model.mesh.setRecombine(2, surf_rock)
+
+    return {
```
brunomariz, line 402: @Olender , I thought I saw some dataclasses in spyro somewhere, but I'm not sure. Are you making this transition or mostly using dicts?

In any case, I wanted to point out that I believe dataclasses improve usability because other programmers can see the fields in dataclass objects more easily than dicts, which can have any keys. This also makes dataclasses less error-prone, because editors can autocomplete fields in a dataclass, but usually not dict keys.

spyro/meshing/meshing_utils.py
brunomariz: My understanding is that the `spyro/io/` folder is the appropriate place for IO related functions, so it would just be a matter of moving IO-related things there, but I have to ask @Olender to confirm this information.

spyro/meshing/meshing_gmsh2d.py
rasjr1305: No special reason to be defined as inner function, it can be moved to the meshing utils folder, I will make this adjustment.

spyro/meshing/meshing_gmsh2d.py
rasjr1305: It is a great idea, however in this case, they are not sepparated in functions to follow the previous already made code, as is done with the SeismicMesh and the Firedrake mesh. We can make a full refactor of the meshing functions making the division into functions in a refactor pr ?

spyro/meshing/meshing_gmsh2d.py
rasjr1305: Same as above, this can be moved to a utils or tools folder

spyro/io/material_properties_io.py
```diff
@@ -233,11 +233,16 @@ def _initialize_material_property_from_file(wave, property_name, from_file, V):
     element_family = V.ufl_element().family()
     element_degree = V.ufl_element().degree()
 
-    print(f"Assigning {property_name} from file {from_file}",
+    source_label = from_file if isinstance(from_file, str) else "grid velocity data"
```
ruansava, line 236: What are the other options?

spyro/io/material_properties_io.py
```diff
@@ -233,11 +233,16 @@ def _initialize_material_property_from_file(wave, property_name, from_file, V):
...
+    print(f"Assigning {property_name} from {source_label}",
           ("in the same" if element_family == original_family
            and element_degree == original_degree else "in another"),
           f"function space: {element_family} {element_degree}.", flush=True)
 
+    if isinstance(from_file, dict):
```
ruansava, line 242: Please, add a comment related the interpolation operation

spyro/io/segy_io.py
```diff
@@ -0,0 +1,114 @@
+import numpy as np
+import segyio
+from io import BytesIO
+from ..io import write_function_to_grid
+import matplotlib.pyplot as plt
+
+
+def segy_to_png(segy_filename, output_file='debug.png', cmap='seismic', vmin=None, vmax=None, dpi=150, flip=True):
```
ruansava: I prefer to add a flag like "show" = True, and output file=None or default name

spyro/io/segy_io.py
```diff
@@ -0,0 +1,114 @@
...
+        return buf.getvalue()
+
+
+def create_segy_from_grid(velocity, filename, rotate=False):
+    spec = segyio.spec()
+
+    if rotate is False:
+        velocity = np.flipud(velocity.T)
+
+        spec.sorting = 2  # not sure what this means
```
ruansava, line 95: I've been seeing this comment ever since I discovered Spyro :v
It's a UCO: Unidentified Commented Object

spyro/io/segy_io.py
```diff
@@ -0,0 +1,114 @@
...
+        return output_file
+    else:
+        buf = BytesIO()
+        fig.savefig(buf, format='png', bbox_inches='tight', dpi=dpi)
+        plt.close(fig)
+        buf.seek(0)
+        return buf.getvalue()
+
+
+def create_segy_from_grid(velocity, filename, rotate=False):
```
ruansava, line 64: Please, add documentation.

spyro/meshing/firedrake_based_wrappers.py
```diff
@@ -0,0 +1,177 @@
+import firedrake as fire
```
ruansava, line 1: I liked to separate this from meshing_functions.py

spyro/meshing/meshing_functions.py
ruansava: I prefer

`if not self.mesh_parameters.is_complete:`

Do you agree?

spyro/receivers/dirac_delta_projector.py
```diff
@@ -117,7 +118,10 @@ def build_maps(self, order=0):
             self.cellVertices,
             self.cellNodeMaps,
         ) = self.__point_locator()
-        self.cell_tabulations = self.__func_build_cell_tabulations(order)
+        if self.use_vertex_only_mesh is False:
```
ruansava, line 121: The same for booleans in conditionals
`if not self.use_vertex_only_mesh:`

Or maybe
`self.cell_tabulations = None if self.use_vertex_only_mesh else self.__func_build_cell_tabulations(order)`

testing_mesher_with_velocity_grid.py
```diff
@@ -0,0 +1,112 @@
+import spyro
+import matplotlib.pyplot as plt
+from spyro.meshing.meshing_parameters import MeshingParameters
+from spyro.meshing.meshing_functions import AutomaticMesh
+import firedrake as fire
+
+
```
ruansava: Add more documentation to defs

propagating_mesh_with_conditional_vp.py
```diff
@@ -0,0 +1,133 @@
+import firedrake as fire
```
ruansava: Add more documentation

propagating_mesh_with_avenir_segy.py
```diff
@@ -0,0 +1,71 @@
+import spyro
```
ruansava: Add more documentation

tests/on_one_core/test_scalar_conditional_to_grid.py
```diff
@@ -0,0 +1,92 @@
+import numpy as np
```
ruansava, line 1: Add mode documentation

tests/on_one_core/test_gmsh2d.py
```diff
@@ -0,0 +1,219 @@
+import math
```
ruansava, line 1: Addo documentation

spyro/receivers/dirac_delta_projector.py
ruansava, line 1: Some defs need documentation

ruansava (changes requested):
I felt there was a lack of comments throughout the code. Sometimes they help follow its logic.

spyro/meshing/meshing_gmsh2d.py
rasjr1305: Moved this function to the meshing utils

spyro/meshing/meshing_gmsh2d.py
rasjr1305: Moved this function to meshing utils

.github/workflows/backup-tags.yaml
Olender, line 1: This should not have been deleted. I'm adding it back

spyro/abc/rec_lay.py
Olender, line 1: Why was this deleted?

spyro/habc/habc.py
Olender, line 1: WHats going on with these abc related changes?

Olender:
I figured out the problem @rasjr1305 . WHen you tried updating your branch with main, you used an old/wrong main in [Merge remote-tracking branch 'origin/main' into romildo/2d_meshing](https://github.com/NDF-Poli-USP/spyro/pull/270/commits/1254265b036c36d8413bd177f5035fc001bad1d7)
[1254265](https://github.com/NDF-Poli-USP/spyro/pull/270/commits/1254265b036c36d8413bd177f5035fc001bad1d7).

Olender:
This usually isn't a problem (normally just a correct merge main would fix it), but you either resolved conflicts incorrectly or force-pushed

Olender:
After a lot of investigation, this branch is currently in a **mixed and inconsistent state relative to `origin/main`**.

It appears that during earlier merges (notably `1254265b` ), parts of the worng `main` were integrated with conflict resolutions that unintentionally dropped or replaced several updates from `main`, including newly added files and updated implementations. As a result, the branch now:

* deletes files that exist in `main` (like out tags, Ruben habc to abc refactoring)
* contains older versions of several modules compared to `main`
* includes structural refactors (e.g., `abc → habc`) that are not consistently aligned with the current `main` state
* shows a large diff against `origin/main` that is not attributable to a single coherent feature set (there are changes that shouldn't be in this branch)

Given this level of divergence, continuing to fix this branch incrementally would be error-prone and difficult to review safely.

### Suggested resolution 1

Closing this PR and replacing it with a clean branch created from the latest `origin/main`, then re-applying only the intended feature commits via cherry-pick.

This will ensure:

* a clean history
* no accidental deletions from `main`
* a reviewable and minimal diff focused only on the intended changes

If needed, I can help reconstruct the branch from the relevant commits.

### Suggested resolution 2
Revert the branch to commit 559e4f3c (pre-merge clean state) and then cherry-pick the relevant changes introduced after that point

I prefer option 2, but I will wait on your @rasjr1305 response

rasjr1305:
> After a lot of investigation, this branch is currently in a **mixed and inconsistent state relative to `origin/main`**.
> 
> It appears that during earlier merges (notably `1254265b` ), parts of the worng `main` were integrated with conflict resolutions that unintentionally dropped or replaced several updates from `main`, including newly added files and updated implementations. As a result, the branch now:
> 
> * deletes files that exist in `main` (like out tags, Ruben habc to abc refactoring)
> * contains older versions of several modules compared to `main`
> * includes structural refactors (e.g., `abc → habc`) that are not consistently aligned with the current `main` state
> * shows a large diff against `origin/main` that is not attributable to a single coherent feature set (there are changes that shouldn't be in this branch)
> 
> Given this level of divergence, continuing to fix this branch incrementally would be error-prone and difficult to review safely.
> 
> ### Suggested resolution 1
> Closing this PR and replacing it with a clean branch created from the latest `origin/main`, then re-applying only the intended feature commits via cherry-pick.
> 
> This will ensure:
> 
> * a clean history
> * no accidental deletions from `main`
> * a reviewable and minimal diff focused only on the intended changes
> 
> If needed, I can help reconstruct the branch from the relevant commits.
> 
> ### Suggested resolution 2
> Revert the branch to commit [559e4f3](https://github.com/NDF-Poli-USP/spyro/commit/559e4f3c45640d9300f5293cec4c5ea706ab4a3f) (pre-merge clean state) and then cherry-pick the relevant changes introduced after that point
> 
> I prefer option 2, but I will wait on your @rasjr1305 response

Yes lets revert to 559e4f3 it was the last safe clean commit. I believe creating from scratch might make something left behind. It seems I merged the "origin/main" while I was supposed to merge the "main" branch, I will have more attention to that in the next commits.

Olender:
> > After a lot of investigation, this branch is currently in a **mixed and inconsistent state relative to `origin/main`**.
> > It appears that during earlier merges (notably `1254265b` ), parts of the worng `main` were integrated with conflict resolutions that unintentionally dropped or replaced several updates from `main`, including newly added files and updated implementations. As a result, the branch now:
> > 
> > * deletes files that exist in `main` (like out tags, Ruben habc to abc refactoring)
> > * contains older versions of several modules compared to `main`
> > * includes structural refactors (e.g., `abc → habc`) that are not consistently aligned with the current `main` state
> > * shows a large diff against `origin/main` that is not attributable to a single coherent feature set (there are changes that shouldn't be in this branch)
> > 
> > Given this level of divergence, continuing to fix this branch incrementally would be error-prone and difficult to review safely.
> > ### Suggested resolution 1
> > Closing this PR and replacing it with a clean branch created from the latest `origin/main`, then re-applying only the intended feature commits via cherry-pick.
> > This will ensure:
> > 
> > * a clean history
> > * no accidental deletions from `main`
> > * a reviewable and minimal diff focused only on the intended changes
> > 
> > If needed, I can help reconstruct the branch from the relevant commits.
> > ### Suggested resolution 2
> > Revert the branch to commit [559e4f3](https://github.com/NDF-Poli-USP/spyro/commit/559e4f3c45640d9300f5293cec4c5ea706ab4a3f) (pre-merge clean state) and then cherry-pick the relevant changes introduced after that point
> > I prefer option 2, but I will wait on your @rasjr1305 response
> 
> Yes lets revert to [559e4f3](https://github.com/NDF-Poli-USP/spyro/commit/559e4f3c45640d9300f5293cec4c5ea706ab4a3f) it was the last safe clean commit. I believe creating from scratch might make something left behind. It seems I merged the "origin/main" while I was supposed to merge the "main" branch, I will have more attention to that in the next commits.

Done

spyro/solvers/backward_time_integration.py
```diff
@@ -139,6 +140,12 @@ def _build_gradient_solver(wave_obj: Wave, mask_available: bool) -> tuple[
     --------
     grad_solver, forward_field, uadj, gradi
     """
+    if wave_obj.use_vertex_only_mesh and wave_obj.automatic_adjoint is False:
```
Olender, line 148: Mega gambiarra, since this error showed up on this PR but is unrelated to it. I've opened an issue #313

spyro/meshing/meshing_gmsh2d.py
Olender: We are making this transition, but currently no one is assigned to it. If this is high priority for total I can make it, since it should no take longer than a week

spyro/io/segy_io.py
Olender: There's a default output_file as a kwarg. I've added a show=False per your suggestion

spyro/io/segy_io.py
Olender: Yes. I'm still not sure what it means

spyro/io/segy_io.py
Olender: Added

spyro/meshing/firedrake_based_wrappers.py
Olender: Yes, I want to make meshing_functions shorter and more organized

spyro/meshing/meshing_functions.py
Olender: I agree. I think it has greatly reduced in size

spyro/meshing/meshing_functions.py
Olender: I really like if is_complete is False, I find it more readable

Olender: I think pep8 favors your approach, I need to check

spyro/meshing/meshing_gmsh2d.py
Olender: Let's open an issue to think about this and leave it to a future PR

spyro/meshing/meshing_parameters.py
Olender: Lets open an issue for this and leave it for a future PR

spyro/meshing/meshing_utils.py
Olender: Yes, its just a matter of moving the method

testing_mesher_with_velocity_grid.py
Olender: These are just temporary functions and tests in the root directory. We will clean them up before merging

spyro/meshing/meshing_gmsh2d.py
rasjr1305: Yes once the meshing features are all working, we can refactor in a new PR.

.github/workflows/backup-tags.yaml
rasjr1305: This was related to the wrong "origin/main" merge.

spyro/abc/rec_lay.py
rasjr1305: This was related to the wrong "origin/main" merge.

spyro/habc/habc.py
rasjr1305: This was related to the wrong "origin/main" merge.

spyro/solvers/backward_time_integration.py
rasjr1305: I believe the correct method is now being used.

rasjr1305:
I believe once the features are accepted a tutorial should be created, I will be working on it

spyro/meshing/meshing_functions.py
```diff
@@ -449,15 +449,43 @@ def create_gmsh_2D_mesh(self):
...
+            if vp.ndim != 2:
+                raise ValueError(
+                    "Grid velocity_model['vp_values'] must be a 2-D array, "
+                    f"but received shape {vp.shape}."
+                )
+
+            vp_for_gmsh = np.ascontiguousarray(vp[::-1, :])
+
+            filename = "tmp_velocity_model.segy"
+            create_segy_from_grid(vp_for_gmsh, filename)
```
Olender, line 480: Something weird is going on for this to be needed. lets leave it for the next PR and open an issue

spyro/meshing/meshing_functions.py
Olender: I really think this should be called hmin here and on the meshing_parameters

rasjr1305: Can we do this on the next PR ?

spyro/meshing/meshing_functions.py
rasjr1305: Both functions, the direct segy and the vp_grid creation have different axis order. For the vp_grid method I found that a clean solution was to test both methods and invert for vp_grid.

spyro/meshing/meshing_functions.py
rasjr1305: Solved

spyro/meshing/meshing_functions.py
rasjr1305: Solved !

spyro/meshing/meshing_functions.py
rasjr1305: Solved

spyro/meshing/meshing_parameters.py
rasjr1305: Now the model accepts a vp_grid generated by the Spyro internal functions, or it can accept an external .segy file and directly convert to an interpolation function

spyro/meshing/meshing_parameters.py
rasjr1305: Now the model accepts a vp_grid generated by the Spyro internal functions, or it can accept an external .segy file and directly convert to an interpolation function

spyro/meshing/meshing_parameters.py
rasjr1305: Now the grading follows the same as Seismic Mesh, 0.1 means small grading (close to 10% of the grid spacing nz nx ny) 0.9 means high grading (90% of the grid spacing). Small grading means medium quality elements but accurate geometry, high grading can generate high quality elements at the cost of geometry conformity

spyro/meshing/meshing_parameters.py
rasjr1305: To find the water interface, the mesher uses all the values in the vp or vs that have the "water_search_value". To detect the interface, the mesher goes from the surface to the bottom of the seafloor testing if the vp or vs is still a water_search_value. When the velocity changes to solid, the seafloor point is found.

spyro/meshing/meshing_parameters.py
rasjr1305: We will do this in a next PR, centralize padding extensions

spyro/meshing/meshing_parameters.py
rasjr1305: ok!

tests/on_one_core/test_gmsh2d_conditional_adaptation.py
```diff
@@ -0,0 +1,263 @@
...
+    dictionary["mesh"] = {
+        "length_z": length_z,  # depth in km - always positive
+        "length_x": length_x,  # width in km - always positive
+        "length_y": 0.0,  # thickness in km - always positive
+        "mesh_type": mesh_type,
+        "velocity_model": vp_grid,
+        "cells_per_wavelength": cells_per_wavelength,
+        "padding_type": None,  # Padding types "rectangular" "hyperelliptical" None
+        "hmin_segy": 0.0,  # Minimum Element size for segy, will apply if higher than function minimum
+        "grade": None,  # function grading for smooth element transition, None = no smooth, 0.9 = high smooth, 0.1 = low smooth
```
rasjr1305, line 191: The Savitzky Golay filter seems to have changed its formulation across scipy versions. So for this test to work, the same scipy version should be local and on the test server. With grade == None this test also work, because it does not uses the Savitzky Golay filter.

tests/on_one_core/test_gmsh2d_conditional_adaptation.py
```diff
@@ -0,0 +1,263 @@
...
+    due to potentional grading and smoothing effects and 1% for outside
+    region.
+    """
+    quadrilateral = False
+    mesh_type = "gmsh_mesh"
+    cells_per_wavelength = 2.6
+    dt = 0.001
+    length_z = 2.0
+    length_x = 2.5
+    vp_grid, cond = create_grid_based_velocity_field(0.02, length_z, length_x)
```
rasjr1305, line 164: It seems the firedrake 2026 version is using a new function to create the rectangle mesh:

    plex = PETSc.DMPlex().createBoxMesh(
        (nx, ny),
        lower=(originX, originY),
        upper=(Lx, Ly),
        simplex=False,
        interpolate=True,
        comm=comm
    )

this is different from 2025 version:

    return TensorRectangleMesh(
        xcoords,
        ycoords,
        quadrilateral=quadrilateral,
        reorder=reorder,
...
    )

I believe this change is creating an interpolation in the vp grid, changing the segy velocity model. This subtle change is making Gmsh create 2 different meshes with different number of elements. This is another reason the test is failing.

Olender: lets open an issue related to this

spyro/io/material_properties_io.py
Olender: Added Issue #320

spyro/meshing/meshing_utils.py
Olender: I've opened issue #321

spyro/receivers/dirac_delta_projector.py
Olender: Yes, but these methods here suffered almost no change from main and are not the subject of this PR. They have existing documentation in the docs branch

tests/on_one_core/test_gmsh2d.py
Olender: @rasjr1305, could you add some? This is also a good place to document with vs code copilot. You could just ask it to document following numpy style docs and revise the output if you want. I don't like AI code, but I do like human-revised AI documentation

tests/on_one_core/test_gmsh2d_conditional_adaptation.py
Olender: I've added the pytest marker. COuld you open the issue? It's really simple, justo go to https://github.com/NDF-Poli-USP/spyro/issues/ click on ´new issue´, describe it there and link the issue here (by typing in a reply Issue #____)

propagating_mesh_with_avenir_segy.py
Olender: These new files in the root directory aren't going to enter main

tests/on_one_core/test_gmsh2d.py
rasjr1305: Yes, I will make a full tutorial once the Pr is accepted

spyro/meshing/meshing_functions.py
rasjr1305: Done !

spyro/meshing/meshing_functions.py
rasjr1305: Done !

spyro/meshing/meshing_functions.py
rasjr1305: Done !

spyro/meshing/meshing_functions.py
rasjr1305: Reduced the size.

spyro/meshing/meshing_utils.py
rasjr1305: Done !

spyro/io/material_properties_io.py
rasjr1305: I will create a full tutorial once the PR is accepted

spyro/io/material_properties_io.py
rasjr1305: I will create a full tutorial once the PR is accepted

tests/on_one_core/test_scalar_conditional_to_grid.py
rasjr1305: I will create a full tutorial once the PR is accepted

Olender:
All checks passed!

Olender:
I vote to merge it and do the other modifications (such as the tutorials) in other PRs asn mentioned. I will cleanup the files I've added for debugging in root and approve. However, the other reviewers who requested changes still need to give their approval

Olender (approved):

Olender:
@brunomariz Your review is absolutely valid and appreciated. Normally I'd never dismiss a review, but since you're in France working with our sponsor, Total, for the next couple of months, we really shouldn't be dragging you back into our day-to-day code reviews. 

I'm dismissing it only so we can merge this PR and deal with things left over in other PRs. Thanks for taking the time to review it and for the valuable comments!

spyro/io/basicio.py
```diff
@@ -314,9 +313,9 @@ def write_function_to_grid(function, V, grid_spacing, buffer=False):
         Interpolated values on grid points
     """
     # get DoF coordinates
-    m = V.ufl_domain()
-    W = fire.VectorFunctionSpace(m, V.ufl_element())
-    coords = fire.assemble(fire.interpolate(m.coordinates, W))
+    mesh = V.ufl_domain()
+    W = fire.VectorFunctionSpace(mesh, V.ufl_element())
```
ruansava, line 329: Centralize the creation of function spaces by using `spyro.domains.space.create_function_space` method

spyro/io/basicio.py
```diff
@@ -314,9 +313,9 @@ def write_function_to_grid(function, V, grid_spacing, buffer=False):
...
+    coords = fire.assemble(fire.interpolate(mesh.coordinates, W))
```
ruansava, line 330: Centralize extraction of node coordinates in ufl or array/tuple form and getting max/min coordinates by using methods in `spyro.meshing.meshing_operations`

ruansava (changes requested):
After these changes, you can merge

spyro/io/basicio.py
Olender: This is being fixed in PR #323

ruansava (approved):

Olender:
I've updated it with main and resolved conflicts. Let's see if it passes and then auto-merge will do its thing

PR #280, opened 2026-04-22
Refactor HABC mesh and improve AutomaticMesh

ruansava:
This pull request primarily focuses on improving naming consistency, mesh creation, and mesh parameter handling in the HABC and meshing modules. The main changes include standardizing the variable name for domain dimensions, refactoring mesh generation logic for clarity and correctness, and enhancing the Firedrake mesh creation functions to handle padding, origin alignment, and distributed meshes more robustly.

### DIFF
--- a/spyro/meshing/meshing_functions.py
+++ b/spyro/meshing/meshing_functions.py
@@ -612,16 +612,14 @@ def RectangleMesh(nx, ny, length_x, length_y, pad=None, comm=None, quadrilateral
     else:
         pad = 0
 
-    dist_par = {"overlap_type": (fire.DistributedMeshOverlapType.NONE, 0)}
     if comm is None:
         mesh = fire.RectangleMesh(nx, ny, length_x, length_y,
-                                  quadrilateral=quadrilateral,
-                                  distribution_parameters=dist_par)
+                                  quadrilateral=quadrilateral)
     else:
         mesh = fire.RectangleMesh(nx, ny, length_x, length_y,
-                                  quadrilateral=quadrilateral,
-                                  distribution_parameters=dist_par,
-                                  comm=comm)
+                                  quadrilateral=quadrilateral, comm=comm)
+
+    # Adjusting to Spyro's reference system (z, x) with origin at (0, 0)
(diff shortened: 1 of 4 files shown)


spyro/meshing/meshing_functions.py
```diff
@@ -641,15 +717,25 @@ def BoxMesh(nx, ny, nz, length_x, length_y, length_z, pad=None, quadrilateral=Fa
         length_z += 2 * pad
     else:
         pad = 0
+
+    dist_par = {"overlap_type": (fire.DistributedMeshOverlapType.NONE, 0)}
```
Olender: Essa linha é na leitura de malhas

spyro/meshing/meshing_functions.py
```diff
@@ -186,19 +217,60 @@ def create_firedrake_mesh(self):
...
+
+        Returns
+        -------
+        discretization : `tuple`
+            Number of elements in each dimension of the mesh.
+            Structure: (nz, nx) for 2D and (nz, nx, ny) for 3D
+        """
+
+        # Compute the edge length if there is no one
+        if self.edge_length is None and self.cpw is not None:
```
rasjr1305, line 251: Here if both are None I think there will be a division by None in the next lines

ruansava: In [meshing_parameters.py](https://github.com/NDF-Poli-USP/spyro/blob/main/spyro/meshing/meshing_parameters.py), these parameters are defined as an XOR operation, then if self.edge_length exists self.cpw does not exist, and vice versa.

spyro/meshing/meshing_functions.py
ruansava: I deleted all the lines with this option.

Olender (approved):

spyro/meshing/meshing_functions.py
Olender: It is possible that both are none. In that case meshing_parameters has the is_incomplete set to True. However, I really want to approve this PR already, so I am just going raise a valueerror here if both are none.

PR #287, opened 2026-04-27
DO NOT SQUASH: Ruben/boundary data acquisition

ruansava:
Improvements to the Eikonal HABC module, including more robust boundary identification and node extraction, and updates docstrings to reflect these changes.


spyro/meshing/meshing_habc.py
```diff
@@ -181,33 +168,32 @@ def original_boundary_data(self):
...
+        all_bnd_nodes = []
+        for (bnd_ids, status) in self.mesh_parameters.boundary_nodes_ids.values():
+            if status:
+                all_bnd_nodes.append(bnd_ids)
+        all_bnd_nodes = np.unique(np.concatenate(all_bnd_nodes))
+        coord_msh = self.mesh_original.coordinates.dat.data_with_halos
+        coord_bnd = node_positions[all_bnd_nodes, :]
+        msh_view = coord_msh.view([('', coord_msh.dtype)] * coord_msh.shape[1])
+        bnd_view = coord_bnd.view([('', coord_bnd.dtype)] * coord_bnd.shape[1])
+        mask_boundary = np.where(np.isin(msh_view, bnd_view))[0]
```
rasjr1305: Here would it be safer to use the kdtree ? Like:

from scipy.spatial import cKDTree
tree = cKDTree(coord_msh)
distances, queried_indices = tree.query(coord_bnd, distance_upper_bound=1e-8)
mask_boundary = queried_indices[queried_indices < len(coord_msh)]

spyro/solvers/eikonal/eikonal_eq.py
```diff
@@ -212,10 +212,9 @@ def define_int_dom(self, V):
         '''
 
         if self.ele_type == 'CG':
-            dx = fire.dx  # At least: degree=2*self.p_eik
-        elif self.ele_type == 'KMV':  # ToDo - Can I use quadrature.py?
-            quad_rule = finat.quadrature.make_quadrature(
-                V.finat_element.cell, self.p_eik, self.ele_type)
+            dx = fire.dx  # At least: degree = 2 * p_eik
```
rasjr1305: There is only a missing test coverage for these 3 lines

rasjr1305 (commented):
Only check the boundary nodes finding logic, to see if its safe for curved boundaries

spyro/solvers/eikonal/eikonal_eq.py
ruansava: I'm going to include this option in the test

spyro/meshing/meshing_habc.py
ruansava: You are right, we can use trees that can be parallelized

ruansava:
> Only check the boundary nodes finding logic, to see if its safe for curved boundaries
Curved boundaries have another logic implemented in layer_boundary_data() method from [meshing_habc.py](https://github.com/NDF-Poli-USP/spyro/blob/main/spyro/meshing/meshing_habc.py)

spyro/habc/eik.py
```diff
@@ -17,13 +18,13 @@ class HABC_Eikonal(Eikonal_Modeling):
...
-    bnds : `tuple` of 'arrays'
-        Mesh node indices on boundaries of the domain. Structure:
-        - (left_boundary, right_boundary, bottom_boundary) for 2D
-        - (left_boundary, right_boundary, bottom_boundary,
-            left_bnd_y, right_bnd_y) for 3D
     bcs_eik : `list`
         Dirichlet BCs for eikonal
+    boundaries : `tuple`
+        Tuple containing the boundary boolean labels for applying absorbing BCs.
+        - (absorb_top, absorb_bottom, absorb_right, absorb_left) for 2D
```
Olender, line 25: I like the added absorb_top option too

spyro/habc/eik.py
```diff
@@ -194,57 +212,51 @@ def ident_crit_eik(self):
...
             # Identifying propagation speed at critical point
-            c_bnd = np.float64(self.c.at(pt_cr).item())
+            c_bnd = np.float64(self.c.at(pnt_crit).item())
 
             # Print critical point coordinates
-            print((eik_str + pnt_str).format(bnd_str, 1e3 * eikmin, *pt_cr))
+            pnt_str = "at (in km): ({2:3.3f}, {3:3.3f})"
+            if self.dimension == 3:  # 3D
+                pnt_str = pnt_str[:-1] + ", {4:3.3f})"
+            print((eik_str + pnt_str).format(bnd_str, 1e3 * eikmin, *pnt_crit))
```
Olender, line 248: We should open an issue to test the eik implementation in parallel with multiple shots. This print without a mpi communicator reminded me of that

spyro/meshing/meshing_habc.py
```diff
@@ -231,24 +219,38 @@ def properties_eik_mesh(self, p_usu=None, ele_type='CG', f_est=0.03):
...
+
         # Setting the properties of the mesh used to solve the Eikonal equation
         self.ele_type_eik = ele_type
         self.p_eik = self.degree if p_usu is None else p_usu
-        self.funct_space_eik = fire.FunctionSpace(self.mesh,
-                                                  self.ele_type_eik,
-                                                  self.p_eik)
+
+        # Function space for the Eikonal modeling
+        if self.quadrilateral and self.ele_type_eik == 'KMV':
```
Olender: This is a bit weird. Why would we have a quadrilateral=True on a simplicial only element type like KMV?

Olender (commented):
I'm good with this going into main, and like how its structured. My only observations would be:
- To open an issue related to parallelism in the eikonal implementation;
- We should figure out why an impossible element might be coming in to one of the methods (quad and kmv), this could be an indication of a bigger problem

spyro/meshing/meshing_habc.py
ruansava: In fact, it's derived from the old implementation. I haven't understood the difference between KMV and spectral elements. I believe that the option should be 'lumped' or 'consistent' as in the damping profile for HABC.

spyro/habc/eik.py
ruansava: We have two options: an independent eikonal for each shot, or solve the eikonal with all sources simultaneously. This means each shot it's a Dirichlet BC inside domain (you gave me the implementation for this BC). The code should support both cases

spyro/meshing/meshing_habc.py
Olender: What do you mean by consistent?

ruansava: That is, the elementar matrices are not diagonal

Olender: So spectral elements is a very broad definition (that has to do with spectral convergence). The 'spectral' variant in Firedrake is just a variant on quadrilateral 'CG' elements that forces them to use gauss-lobatto-legendre nodal values. However, those elements are still 'cg' since they have the same function space em any other 'cg' firedrake element with the same p. The only difference for fire.FunctionSpace(mesh, "CG",4, variant='spectral'), is that the nodes are at GLL points. However, it is still the same space as fire.FunctionSpace(mesh, "CG",4), just with better properties. 

'KMV'degree p is something completly different, its the mass-lumped simplicial (triangular or tetrahedral) elements. The space isn't the same as 'CG' degree p, since we have interior bubble functions from higher-order spaces (varying from p+1 to p+2).

Olender: I agree with a 'lumped' definition. However, that doesn't make sense to me in the eikonal, since not one of the elements we use are lumped for the eikonal equation

Olender:
```suggestion
        if self.quadrilateral:
```

ruansava: I am testing with this. Not committed yet:
        # Function space for the Eikonal modeling
        if ele_type == 'consistent':
            self.funct_space_eik = create_function_space(self.mesh, 'CG', self.p_eik)

        if ele_type == 'lumped':
            method = 'spectral_quadrilateral' if self.quadrilateral \
                else 'mass_lumped_triangle'
            degree = min(self.p_eik, 4 if self.dimension == 2 else 3)
            self.funct_space_eik = create_function_space(self.mesh, method, degree)

Olender: I still don't understand why you would want to call elements 'consistent' here. This place defines the FunctionSpace and not the integration. Also to be honest 'lumped' gives the wrong idea since they are not lumped for the eikonal equation

Olender: What if we just remove this altogether and just use CG with or without spectral?

Olender: Just to be clear, so I understand you definition: by consistent you mean "exact variational integration"?

ruansava: It is because of the construction of the function space. For different elements and methods, we need to set different degrees. Initially, I used only "CG" elements, but then I had to compare with KMV and spectral quads. For this, I am giving support for both cases. However, CG is the default because the degrees and computational cost for this problem are lower. I adopted the nomenclature used in [damp_profile.py](https://github.com/NDF-Poli-USP/spyro/blob/main/spyro/habc/damp_profile.py), which we had already reviewed under the same idea some time ago

ruansava: > Just to be clear, so I understand you definition: by consistent you mean "exact variational integration"?

Yeah!

Olender: Let me just make a brief observation of my reasoning related to the spyro wave variant nomenclature (that might be valid for wave equations, not here):
- So everything comes from Firedrake, which bases its definition on the Ciarlet FEM definition. Ciarlet defines a finite element as a triple (element domain, space of shape functions, set of DoFs).
- If we were to use the Ciarlet definition for the elements used in spyro (Quad spectral or triangular ML):
	
	For mass-lumped triangular:
	(triangle, P, N), where P = Pk (+) B (standard polynomials of degree <=k with the added bubble) and N are our nodes
	
	For spectral quads:
	(quadrilateral, P, N), where P are the tensor product polynomials and N are our nodes

Right now, our elements themselves are not “consistent” or “lumped”,in the Ciarlet sense they are simply (K,P,N).
Consistency or lumping only arises at the level of the assembled operator, i.e., depending on the quadrature used to evaluate the variational forms.

For wave propagation, we explicitly choose a quadrature that produces a diagonal mass matrix (the “lumped” variant).

For eikonal, you do not enforce this, and instead use standard integration.

So “consistent” here really refers to the integration strategy (exact or sufficiently accurate quadrature), not to the element itself.

However, for eikonal we are not lumping. Using any of the above elements (triangle, P, N), or (quadrilateral, P, N) without specifiying a quadrature rule will give us a 'consistent' element by default.

So calling “non-GLL quads” consistent elements is misleading, since consistency is not an intrinsic property of the element but of how the bilinear forms are integrated

Olender: Also calling any element 'lumped' here is also misleading, since they are not lumped

Olender: These name are all a bit crazy since 'spectral' means something completely different in the FEM communities and in the geophysical communities

ruansava: What should be the labels to differentiate the types of elements?

ruansava: I am going to update the code with the new test.

Olender: If you want to use the 'consistent' nomenclature it would make more sense to name the other elements (with underintegration) 'under-integrated'

ruansava: "full" and "reduced" in the sense of integration?

Olender: Also spectral quadrilateral elements remain spectral independently of the quadrature choice: using exact (or sufficiently accurate) quadrature gives the standard consistent Galerkin formulation or under-integrated GLL quad points

ruansava: I got!

Olender: I think github is buggy today, since I can't see all of your responses on the github website, but have now seen them on my vs code github extension. I apologize if it seemed that I was ignoring you responses. This was not intentional

spyro/habc/damp_profile.py
```diff
@@ -245,7 +245,7 @@ def Zi(p, alpha, ele_type):
                 alpha : `float`
                     Ratio between the representative mesh dimensions
                 ele_type : `string`
-                    Element type. 'consistent' or 'lumped'
+                    Finite element type. 'consistent' or 'lumped'
```
Olender, line 248: quadrature type?

Olender: consistent and underintegrate

Olender (approved):

spyro/meshing/meshing_habc.py
ruansava: Done!

spyro/meshing/meshing_habc.py
```diff
@@ -181,11 +181,14 @@ def original_boundary_data(self):
         all_bnd_nodes = np.unique(np.concatenate(all_bnd_nodes))
         coord_msh = self.mesh_original.coordinates.dat.data_with_halos
         coord_bnd = node_positions[all_bnd_nodes, :]
-        msh_view = coord_msh.view([('', coord_msh.dtype)] * coord_msh.shape[1])
-        bnd_view = coord_bnd.view([('', coord_bnd.dtype)] * coord_bnd.shape[1])
-        mask_boundary = np.where(np.isin(msh_view, bnd_view))[0]
         self.bnd_nodes = coord_bnd
 
+        # Identify the boundary nodes
+        tree = cKDTree(coord_msh)
```
rasjr1305, line 187: This is faster than the previous solution !

rasjr1305 (approved):
The asked changes have been implemented, improving the initial commit with better boundary detection and testing, so I will approve now

spyro/meshing/meshing_habc.py
ruansava: Yeah!

PR #302, opened 2026-05-12
Olender romildo/gmshstuff

Olender:
Small changes:
- I took out the default water value. If we add one it should be in only one place. The first time it shows up is in spyro.meshing.meshing_paramters.py line 356. I've left default as None, but we can change it if you want. However, sometimes the segy file is in km/s and sometimes m/s, so a default value can break some things.

-linting

### DIFF
--- a/spyro/meshing/meshing_utils.py
+++ b/spyro/meshing/meshing_utils.py
@@ -83,6 +83,13 @@ def create_sizing_function(
     # Set water velocity if value = 0
     if vp_water is not None:
         vp = np.where(vp == 0, vp_water, vp)
+    else:
+        # If no water velocity provided and model contains zeros, raise error
+        if np.any(vp == 0):
+            raise ValueError(
+                "Velocity model contains zero values (water). Provide 'vp_water' "
+                "to substitute water velocities or preprocess the SEGY file."
+            )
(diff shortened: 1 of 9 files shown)


spyro/meshing/meshing_utils.py
```diff
@@ -86,8 +102,7 @@ def create_sizing_function(fname, hmin=None, bbox=None, wl=10, freq=2, pad_type=
         dx = (bbox[3] - bbox[2]) / n_samples
         nnz = int(pad_size_z / dz)
         nnx = int(pad_size_x / dx)
-        print(nnx, nnz, n_samples, n_traces)
-        print(pad_size_z, bbox[0], bbox[2])
+
```
Olender, line 105: I took the print out because of mintrop runs. If you want I can add them back with parallel_print. When we run in mintrop without using parallel_print (or without a flush=True kwarg) these kinds of prints will appear anywhere in the output (even on 1 core)

Olender: So in Mintrop runs (at least the last time I checked), regular `print` statements executed in sequence without `flush=True` may appear in an arbitrary order relative to each other, and will also only show up together with the next `parallel_print` (or any `print(..., flush=True)`) that becomes available.

PR #311, opened 2026-06-11
New native hdf5 converter

rasjr1305:
Added new hdf5 converter and new functions. Improved some user interface when compared to the SeismicMesh version.

New IO functions:

    - read_segy_velocity_model,
    - read_bin_velocity_model,
    - write_velocity_model


spyro/io/basicio.py
```diff
@@ -699,3 +699,407 @@ def saving_source_and_receiver_location_in_csv(model, folder_name=None):
...
+        nx = len(segy.trace)
+        nz = len(segy.samples)
+        vp = np.zeros((nz, nx), dtype=np.float32)
+
+        for i in range(nx):
+            vp[:, i] = segy.trace[i]
+
+    vp = np.flipud(vp)
+
+    return vp, nx, nz
```
Olender:
```suggestion
    nz : int
        Number of samples per trace, corresponding to the z direction.
    nx : int
        Number of traces in the SEG-Y file, corresponding to the x direction.

    Raises
    ------
    ImportError
        If ``segyio`` is not installed.
    """
    with segyio.open(fname, "r", ignore_geometry=True) as segy:
        nx = len(segy.trace)
        nz = len(segy.samples)
        vp = np.zeros((nz, nx), dtype=np.float32)

        for i in range(nx):
            vp[:, i] = segy.trace[i]

    vp = np.flipud(vp)

    return vp, nz, nx
```

spyro/io/basicio.py
```diff
@@ -699,3 +699,407 @@ def saving_source_and_receiver_location_in_csv(model, folder_name=None):
...
+            "axes_order tuple/list must contain either all integers or all strings."
+        )
+
+    raise TypeError(
+        "axes_order must be a string like 'z x y'/'zxy'/'1 0 2'/'102' "
+        "or a tuple/list like (2, 0, 1)."
+    )
+
+
+def read_bin_velocity_model(
```
Olender, line 906: This function only works in 3D or can we use it in 2D with ny = 0? I'm okay with it only working in 3D. However, I think we need to add an observation in the documentation if thats the case

spyro/utils/utils.py
```diff
@@ -730,42 +726,29 @@ def write_hdf5_velocity_model(obj_with_comm, segy_filename):
...
-        segy_filename, ofname=vp_filename
+    vp_filename, _ = os.path.splitext(segy_filename)
+
+    output_filename = write_velocity_model(
+        filename=segy_filename,
+        ofname=vp_filename,
+        model_type="segy",
     )
-    output_filename = vp_filename + ".hdf5"
+
```
Olender, line 751: Yes!

tests/on_one_core/test_io_hdf5.py
```diff
@@ -0,0 +1,392 @@
...
+import pytest
+
+from spyro.io.basicio import (
+    read_bin_velocity_model,
+    read_segy_velocity_model,
+    write_velocity_model,
+    _parse_axes_order,
+)
+
+HAS_SEGYIO = importlib.util.find_spec("segyio") is not None
```
Olender: segyio is a hard dependency, so we don't have to check for it. Spyro should just not work if it isn't installed

Olender:
I love the new functions. I just think they should be moved to the segy_io file (it isn't here, but I've added one in PR #270 ), since basic_io is getting very crowded. However, I think we should merge this into main before moving and move them after PR#270 is in with a quick new PR

spyro/io/basicio.py
Olender: Infelizmente o pessoal da sísmica gosta do alfabeto ao contrário. Geralmente vejo z antes de x ao guardar os dados e ler eles em arrays. Engraçado que essa mesma galera coloca x, z ao plotar figuras. Vai entender.

Seismic folks seem to enjoy reversing the alphabet. It's common to store data in arrays as (z, x) (or [iz, ix]) while plotting and referring to physical coordinates as (x, z)

tests/on_one_core/test_io_hdf5.py
```diff
@@ -0,0 +1,392 @@
...
+    vp, nx, nz = read_segy_velocity_model(str(segy_file))
+    expected = _expected_segy_model(segy_file)
+
+    assert vp.shape == expected.shape
+    assert nx == expected.shape[1]
+    assert nz == expected.shape[0]
+    assert np.array_equal(vp, expected)
+
+
+@pytest.mark.skipif(not HAS_SEGYIO, reason="segyio is not installed")
```
Olender:
```suggestion
```

tests/on_one_core/test_io_hdf5.py
Olender:
```suggestion
```

Olender (changes requested):
Sugeri umas mudanças bem pequenas. Dê uma olhada. Vamos dar merge nisso hoje

spyro/io/basicio.py
rasjr1305: should now work with 3d and 2d if ny = 0

Olender (approved):

PR #314, opened 2026-06-25
Test modal fail

Ig-dolci:
The parametrised modal tests (``homogeneous`` True/False, 2D/3D, and T/Q elements) all write their results under ``output/modal_test{dimension}d``. Under ``pytest -n`` (xdist), several of these items run concurrently in separate worker processes and would otherwise share that single directory. ``rename_folder_habc`` then ``rmtree``/``rename``s a per-case folder that another worker is still writing into, which raises ``FileNotFoundError`` (e.g. on ``RAYLEIGH_cost.txt``). Scoping the folder by the xdist worker id gives concurrent items distinct directories, so they can no longer collide. Falls back to ``main`` when not running under xdist.

These failures can be observed in this [PR CI tests](https://github.com/NDF-Poli-USP/spyro/actions/runs/28136277642/job/83325238229?pr=279)

### DIFF
--- a/tests/on_one_core/test_modal_solvers.py
+++ b/tests/on_one_core/test_modal_solvers.py
@@ -3,7 +3,7 @@ import warnings
 import numpy as np
 import firedrake as fire
 import spyro.habc.habc as habc
-from os import makedirs, path
+from os import makedirs, path, environ
 from spyro.utils.cost import comp_cost
 fire.parameters["loopy"] = {"silenced_warnings": ["v1_scheduler_fallback"]}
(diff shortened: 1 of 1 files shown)


ruansava:
The modal test was refactored in abc_refactoring branch and all tests are passing at this moment. I am waiting for the approving of the respective PR to include these changes

Ig-dolci:
> The modal test was refactored in abc_refactoring branch and all tests are passing at this moment. I am waiting for the approving of the respective PR to include these changes

Ok. Então vou fechar esse mantendo em mente que o abc_refactoring está corrigindo esse erro.

PR #328, opened 2026-07-13
DO NOT SQUASH: PML Refactoring

ruansava:
This pull request refactors and improves the PML module by renaming files, updating class and function names, improving code style, and documentation.

PR #330, opened 2026-07-14
DO NOT SQUASH: Simple accurate vp output

Olender:
TLDR: New method to export scalar fields to segy or png and some io file reorganization

Since we use high-order FEM we cant visualize correctly in paraview. I've introduced a new utility function for exporting scalar fields to segy or png (for paper or report images). Now we can export our data to a more accurate (based on the grid space of the desired sampling) image. Unfortunately, since the parallelism related decorators were all over the place I had to reorganize them as well.

Therefore, this pull request refactors the `spyro/io` module to improve code organization. I just moved the parallelism-related decorators and functions into a new `parallelism_wrappers` module, and updated the imports throughout the codebase. These changes help separate concerns between basic I/O operations and parallelism logic.

### Refactoring and Code Organization

* Moved all parallelism-related decorators and utility functions (such as `ensemble_save`, `ensemble_load`, `ensemble_gradient`, `ensemble_functional`, `delete_tmp_files`, `switch_serial_shot`, `ensemble_shot_record`, `is_owner`, and `ensemble_propagator`) from `basicio.py` to a new `parallelism_wrappers.py` module, and updated imports accordingly throughout the codebase. 

### New Features

* Added a new utility function `export_scalar_field` in `segy_io.py` to export scalar fields to SEG-Y or PNG formats, including logic for intermediate conversion and file validation. (`spyro/io/segy_io.py`,


tests/parallel/test_simple_fwi_with_segy_output.py
```diff
@@ -0,0 +1,230 @@
...
+        "perform_fwi": True,  # switch to true to make a FWI
+        "initial_guess_model_file": None,
+        "shot_record_file": None,
+    }
+
+    shots_filenames="shots/shot_record_"
+
+    # Setting up to run synthetic real problem
+    if load_real_shot is False:
+        fwi_obj = run_forward_real_model(
```
rasjr1305, line 183: Deep in this function, the number of cores is being divided generating an error:

 f"Available cores cannot be divided between sources equally {available_cores}/{model.number_of_sources}."
E                   ValueError: Available cores cannot be divided between sources equally 1/6.

Maybe detect the number of cores them later divide by a possible number, or even just use all the cores

spyro/io/segy_io.py
```diff
@@ -139,6 +140,47 @@ def create_segy(function, V, grid_spacing, filename):
...
+    # if output_filename.lower().endswith((".png", ".segy", ".sgy")):
+    V = function.function_space()
+    if output_filename.lower().endswith((".segy", ".sgy")):
+        create_segy(function, V, grid_spacing, output_filename)
+    elif output_filename.lower().endswith((".png")):
+        segy_filename = Path(output_filename).with_suffix(".segy")
+        create_segy(function, V, grid_spacing, segy_filename)
+        segy_to_png(segy_filename, output_file=output_filename)
+    else:
+        valid_extensions = [".segy", ".sgy", ".png"]
```
ruansava, line 186: Could we create a method named `value_extension_error` (only for importing/exporting files) analogous to `value_parameter_error`?

ruansava (changes requested):

tests/parallel/test_simple_fwi_with_segy_output.py
Olender: This test is not failing. You can verify this in the CI runner: https://github.com/NDF-Poli-USP/spyro/actions/runs/29515081224/job/87685093721?pr=330. It runs and passes with `mpiexec -n 6` without generating any errors.

If it is failing locally in your PC, it is because this is a strictly parallel test (which is also why it is located in the `parallel` folder) and by your error you are running it in serial. In that case, the error message is expected and desired.

This is how automatic parallelism is intended to work in Spyro. It automatically selects the no-approximation, no-additional-memory configuration, where the number of cores has to be a multiple of the number of sources and each source has its own propagation.

If a user wants a different type of parallelism, they can explicitly choose either `spatial`, which uses spatial parallelism only, or `custom`. Spyro supports multiple different types of parallelism, but if the user isn't using an mpi launcher we are failing safely.

The parallelism is transparent to the user: they choose the number of cores they want to use, and the MPI launcher available on their system handles starting the corresponding number of processes

spyro/io/segy_io.py
Olender: I can

Olender: What if we just reuse the existing value_parameter_error method since it is a value parameter error case. Have a look at the latest changes

ruansava: Nice!

ruansava: PR #328 has a new version for complete validation. In the near future, you will be able to implement something cleaner, such as:


```
from ..utils.error_management import value_parameter_error
extension = value_parameter_error("extension_type",  
         Path(output_filename).suffix.lower(), [".segy", ".sgy", ".png"])
V = function.function_space()
segy_filename = Path(output_filename).with_suffix(".segy") \ 
        if extension == ".png" else output_filename
create_segy(function, V, grid_spacing, segy_filename)
if extension == ".png":
         segy_to_png(segy_filename, output_file=output_filename)
```

ruansava (approved):

PR #332, opened 2026-07-22
DO NOT SQUASH: Adding direct segy to grid method

Olender:
TLDR: Added method to convert a SEG-Y file directly into a structured grid dictionary and moved a segy related method from basicio into segy_io

Added a new function, `create_grid_dictionary_from_segy`, to `segy_io.py` to convert a SEG-Y file directly into a structured grid dictionary, including physical dimensions and spacing metadata, together with the relevant test. Also moved the implementation of `read_segy_velocity_model` from `basicio.py` to `segy_io.py`, and updated imports to reflect this.


spyro/io/segy_io.py
```diff
@@ -140,12 +140,98 @@ def create_segy(function, V, grid_spacing, filename):
...
+
+        for i in range(nx):
+            vp[:, i] = segy.trace[i]
+
+    vp = np.flipud(vp)
+
+    return vp, nz, nx
+
+
+def create_grid_dictionary_from_segy(filename, length_z=None, length_x=None):
```
ndf-ti: if both length_z and length_x are required, should we set the default as None?

ndf-ti: You could also add a type hint that these values are float

PNAraujo (approved):

spyro/io/segy_io.py
Olender: Done

ndf-ti (approved):

PR #335, opened 2026-07-23
Add basic models

ndf-ti:
This PR adds models that we can use to do input validation and better organize code. This is currently just a proposal and i need feedback to understand how can we better organize the data structures. For now i created the following data models (please, correct me if they dont make sense to you):

- ExecutionConfig: this model should contain all information required to run a solver. It contains the solver input, parallelism information and the output specification. The main goal of this data model is to decouple computational information (how much cores will be used? which folder will be used to save output?) from the problem specification that will be isolated in the solver_input field.

- ParallelismConfig: This data model will contain any information regarding parallelism. For now it contains only the ParallelismType, shot_ids_per_propagation and the comm.

- OutputSpecification: This data model will contain all information regarding where the output will be saved.

- SolverInput: this model contains the problem specification. It contains the Mesh, Time information, boundaries, equationType, acquisition information and variant (i dont know exactly what the variant is but i left it here for now)

- MeshDefinition: this data model will be equivalent to meshing_parameters. It should contain all information regarding a mesh.

- Mesh: this is the actual mesh object. I saw that there can be some types of these objects so i created a abstract class that we can extend to create new mesh types. I dont know if this is a good approach but i think it can be cleaner than the current approach in meshing_paramters.

- Boundary: This data model will be equivalent to the Read_boundary_layer type. I dont know if it makes more sense for it to be in the mesh or in the SolverInput. It is possible that some fields are missing here and i will probably need the help of you guys.

### DIFF
--- a/spyro/model/boundary.py
+++ b/spyro/model/boundary.py
@@ -1,0 +1,50 @@
+from dataclasses import dataclass
+from enum import Enum
+
+
+class ReferenceFrequency(Enum):
+    BOUNDARY = "boundary"
+    SOURCE = "source"
+
+
+class ExtensionMode(Enum):
+    DRIVEN = "abc_driven"
+    BUILTIN = "builtin"
+
+
+@dataclass
+class BoundaryShape:
+    pass
+
+
+@dataclass
+class RectangularBoundary(BoundaryShape):
+    pass
+
+
+@dataclass
+class HypershapeBoundary(BoundaryShape):
+    degree: int
+    degree_type: str
+
+
+@dataclass
+class Boundary:
+    ref_model: bool
+    reference_freq: ReferenceFrequency
+    extension_mode: ExtensionMode
+    pad_length: float
+    shape: BoundaryShape
+    degree_eikonal: int = 2
+
+    def __post_init__(self):
+        if self.pad_length < 0:
+            raise ValueError("Pad length must be positive")
+
+
+@dataclass
+class PMLBoundary(Boundary):
+    shape: RectangularBoundary
+    max_acoustic_velocity: float
+    exponent: float = 2
+    R: float = 1e-6
--- a/spyro/model/mesh.py
+++ b/spyro/model/mesh.py
@@ -1,0 +1,125 @@
+from dataclasses import dataclass, field
+from abc import ABC, abstractmethod
+from enum import Enum
+import firedrake as fire
+from .. import meshing
+
+
+class MeshType(Enum):
+    FIREDRAKE_MESH = "firedrake_mesh"
+    USER_MESH = "user_mesh"
+    SEISMIC_MESH = "SeismicMesh"
+    FILE = "file"
+    SPYRO_MESH = "spyro_mesh"
+    GMSH_MESH = "gmsh_mesh"
+
+
+class FiniteElementMethod(Enum):
+    MASS_LUMPED_TRIANGLE = "mass_lumped_triangle"
+    DG_TRIANGLE = "DG_triangle"
+    SPECTRAL_QUADRILATERAL = "spectral_quadrilateral"
+    DG_QUATRILATERAL = "DG_quadrilateral"
+    CG_TRIANGLE = "CG_triangle"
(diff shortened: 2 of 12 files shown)


spyro/model/boundary.py
```diff
@@ -0,0 +1,51 @@
...
+        if self.pad_length < 0:
+            raise ValueError("Pad length must be positive")
+
+
+@dataclass
+class PMLBoundary(Boundary):
+    max_acoustic_velocity: float
+    exponent: float = 2
+    R: float = 1e-6
+    cmax: float = 4.7
```
Olender: cmax and max_acoustic_velocity are the same thing in acoustic cases and in almost every elastic case (Vp/Vs is usually around 1.7 in rocks). I think we should have only one of them for now

spyro/model/mesh.py
```diff
@@ -0,0 +1,126 @@
...
+    GMSH_MESH = "gmsh_mesh"
+
+
+class FiniteElementMethod(Enum):
+    MASS_LUMPED_TRIANGLE = "mass_lumped_triangle"
+    DG_TRIANGLE = "DG_triangle"
+    SPECTRAL_QUADRILATERAL = "spectral_quadrilateral"
+    DG_QUATRILATERAL = "DG_quadrilateral"
+    CG_TRIANGLE = "CG_triangle"
+    CG = "CG"
```
Olender: The lone CG method is a leftover from previous input dicitonary keys that no one uses and has been deprecated for 5 yrs. I think we can remove it in this PR, or mark it as a TODO so we can remove support in another PR.

spyro/model/mesh.py
```diff
@@ -0,0 +1,126 @@
...
+    @abstractmethod
+    def get_mesh(self, mesh_definition):
+        pass
+    
+
+@dataclass
+class MeshDefinition:
+    mesh: Mesh
+    domain_dimensions: list[float]
+    negative_z: float
```
Olender: Shouldn't this be a Boolean?

spyro/model/mesh.py
```diff
@@ -0,0 +1,126 @@
...
+            Coordinates of the point to check.
+
+        Raises
+        ------
+        ValueError
+            If the point is outside the mesh domain.
+        """
+        # Avoid changing mesh lengths outside of this
+        if self.negative_z:
+            self.domain_dimensions[0] = -self.domain_dimensions[0]
```
Olender: Will this flip de value repeatedly every time this assert is called for every point?

spyro/model/parallelism.py
```diff
@@ -0,0 +1,28 @@
...
+        if type == ParallelismType.CUSTOM and self.shot_ids_per_propagation is None:
+            raise ValueError(
+                "shot_ids_per_propagation not specified "
+                + "for ParallelismConfig even though type is CUSTOM"
+            )
+
+    def setup(self, number_of_sources):
+        self.comm = utils.mpi_init(self)
+        self.comm.comm.barrier()
+        self.shot_ids_per_propagation = [[i] for i in range(0, number_of_sources)]
```
Olender: This should not be done in custom

spyro/model/solver_input.py
```diff
@@ -0,0 +1,80 @@
...
+    LUMPED = "lumped"
+    EQUISPACED = "equispaced"
+    DG = "DG"
+
+
+@dataclass
+class Acquisition:
+    receiver_locations: np.array
+    source_type: SourceTypes
+    source_locations: np.array  # can this be null?
```
Olender: We could set it to None in the MMS source case

spyro/model/boundary.py
ndf-ti: Thanks!

spyro/model/mesh.py
ndf-ti: yes, i will move this to the __post_init__ as it seems more like an initialization code. Also, why is this flag necessary? We cant pass directly a negative value of Z?

PR #334, opened 2026-07-23
DO NOT SQUASH: moving atributes that only exist in gmsh mesh type

Olender:
This pull request updates the initialization logic in the `MeshingParameters` class to ensure that certain mesh parameters are only set when the mesh type is `gmsh_mesh`. This helps avoid unnecessary, irrelevant or worng default parameter assignments for other mesh types and improves code clarity

PR #336, opened 2026-07-27
Acoustic fwi examples

Olender:
TLDR: Brought back demos and added 2 simple FWI cases to them. Had to debug newer features from previous PRs so they wouldn't interfere with the FWI cases.

Added two new demos: `demos/acoustic_camembert_fwi.py` and `demos/acoustic_layers_fwi.py`. The first Camembert demo takes advantage of the new time interpolator feature. The second layer demo of the newer segy io features and @rasjr1305 first gmsh-based mesher with mesh markings for gradient calculation. To get them to work I had to modify the `_grid_velocity_data_to_source_function` and the `project_grid_velocity_data` methods to accept a mpi communicator and had to stop the gmsh_mesh default parameters from interfering and being a default for all mesh types.

I've also added new tests to test the demos. Since they are quite big they will not become part of our standart test suite and will only run on Sunday. In the future, I plan on adding the notebooks to the same Sunday run


demos/acoustic_layers_fwi.py
```diff
@@ -0,0 +1,269 @@
...
+    layers list. The layers are separated by the values in the z_switch list.
+
+    Parameters
+    ----------
+    z_switch : list of floats
+        List of z values that separate the layers.
+    layers : list of floats
+        List of velocity values for each layer.
+    """
+    if len(z_switch) != (len(layers) - 1):
```
ruansava, line 87: Error methods can help to validate parameters here

spyro/io/interpolators.py
```diff
@@ -0,0 +1,258 @@
...
+        "edge_length_y": grid_spacing_y,
+        "abc_pad_length": grid_velocity_data.get("abc_pad_length"),
+    }
+    source_mesh = AutomaticMesh(
+        MeshingParameters(input_mesh_dictionary=source_mesh_parameters, comm=comm)
+    ).create_mesh()
+
+    source_space = fire.FunctionSpace(source_mesh, "CG", 1)
+    source = fire.Function(source_space)
+    source_coords = source_mesh.coordinates.dat.data
```
ruansava, line 40: There are some methods for mesh operations useful here

spyro/io/interpolators.py
```diff
@@ -0,0 +1,258 @@
...
+    c: Firedrake.Function
+        P-wave seismic velocity interpolated onto the nodes
+        of the finite elements.
+
+    """
+    m = V.ufl_domain()
+
+    add_pad = False
+    if Model.mesh_parameters.abc_pad_length is not None:
+        if Model.mesh_parameters.abc_pad_length > 1e-15:
```
ruansava: Why this value?

spyro/meshing/meshing_parameters.py
```diff
@@ -287,8 +274,24 @@ def __init__(self, input_mesh_dictionary=None, dimension=None, source_frequency=
...
         self.padding_x = self.input_mesh_dictionary.get("padding_x")
         self.padding_z = self.input_mesh_dictionary.get("padding_z")
-        self.h_padding = self.input_mesh_dictionary.get("h_padding", 500.0)
-        self.winslow_implementation = self.input_mesh_dictionary.get("winslow_implementation", "numba")
+
+        # Winslow gmsh only parameters
+        if self.mesh_type == "gmsh_mesh":
+            self.winslow_implementation = self.input_mesh_dictionary.get("winslow_implementation", "numba")
+            self.h_padding = self.input_mesh_dictionary.get("h_padding", 500.0)
+            self.vp_water = self.input_mesh_dictionary.get("vp_water", None)
```
ruansava, line 282: Why is h_padding 500?

simple_fwi.py
```diff
@@ -0,0 +1,196 @@
...
+    if case == "camembert":
+        # Builds the true velocity model based on a conditional
+    
+        center_z = -1.0
+        center_x = 1.0
+        mesh_z = fwi_obj.wave.mesh_z
+        mesh_x = fwi_obj.wave.mesh_x
+        cond = fire.conditional((mesh_z-center_z)**2 + (mesh_x-center_x)**2 < .2**2, 3.0, 2.5)
+    elif case not in supported_cases:
+        return ValueError(f"Case of {case} not part of supported cases: {supported_cases}")
```
ruansava: value_parameter_error()

simple_fwi.py
```diff
@@ -0,0 +1,196 @@
...
+    layers list. The layers are separated by the values in the z_switch list.
+
+    Parameters
+    ----------
+    z_switch : list of floats
+        List of z values that separate the layers.
+    layers : list of floats
+        List of velocity values for each layer.
+    """
+    if len(z_switch) != (len(layers) - 1):
```
ruansava: Add validations

ruansava (changes requested):

spyro/meshing/meshing_parameters.py
Olender: No ideia. I just moved those values out of the way. This is actually already in main because of PR #334 that was already merged (this branch was made on top of that PR's branch before the merge). I don't really want to deal with the default gmsh_mesh mesh type values in these PRs. I just don't want them interfering with the rest of the code

demos/acoustic_layers_fwi.py
Olender: I agree that error methods are useful for parameter validation.

My only hesitation is that these demos are outward-facing examples intended for end users, and in this case a simple `ValueError` is already very clear and easy to understand. I do think the dedicated error methods are a great fit for validation within Spyro itself, where they help keep error handling consistent across the codebase

spyro/io/interpolators.py
Olender: This method was done before the mesh_operations existed. I just moved it to another file for organization purposes in this PR. Thanks for the info, I will update it now

spyro/io/interpolators.py
Olender: Because sometimes the pad is set to 0.0 but it is not really zero and those cases will screw up the fast interpolator

simple_fwi.py
Olender: This file should not exist. I'll delete it

simple_fwi.py
Olender: This file should not exist. I will delete it. It was just an intermediary file I made and based the demos on it

spyro/meshing/meshing_parameters.py
Olender: I've opened an Issue #339

simple_fwi.py
Olender: Done

simple_fwi.py
Olender: Done

demos/acoustic_layers_fwi.py
ruansava: We can try to separate cases for the application of "automated" validation

spyro/io/interpolators.py
ruansava: Ok!

spyro/io/interpolators.py
ruansava: Automated validation :v

ruansava (approved):

spyro/io/interpolators.py
Olender: Do you mean for me to use the MeshOps.extract_node_positions() instead of line 40?

So, here I am only doing three things:
1. Creating a CG1 space,
2. Getting the existing coordinate array.
3. Filling out a function's np.array values.

I really like MeshOps, especially because of its generality, where it can work with arbitrary spaces that are scalar or vector, assembles and interpolated, etc.

However, for these 3 lines of code all of this is unnecessary and might cloud the logic. I would have to substitute line 40, which is pretty clear
from:
```
source_coords = source_mesh.coordinates.dat.data
```
to:
```
mesh_ops = MeshOps(
    domain_dim=(length_z, length_x) if vp_values.ndim == 2
    else (length_z, length_x, length_y),
    dimension=vp_values.ndim,
    func_space_type="scalar",
    comm=comm,
)
source_coords = mesh_ops.extract_node_positions(
    source_mesh,
    source_space,
    output_type="array",
)
```
But there's an extra problem. Extract node postiions does an interpolation and an assemble, which can be costly and would increase runtime here. I currently just read a np array. Since my method is strictly a cg1 method (this is by design so it is closer to the grid) the interpolation can become very expensive (a lot of dofs on the mesh). I prefer to leave it as is.

We could also compromise. I can add a new method to mesh opts, like:
```
def extract_mesh_coordinate_array(self, mesh):
    return mesh.coordinates.dat.data
```

Olender: But I do think it would cloud the logic

spyro/io/interpolators.py
Olender: But I don't want it to fail or pass. I want it to be ignored if small with my boolean

spyro/io/interpolators.py
ruansava: If the current code is the simplest, you can leave it as is.

demos/advanced_simple_fwi_wrapper.py
```diff
@@ -0,0 +1,282 @@
...
+    if len(z_switch) != (len(layers) - 1):
+        raise ValueError(
+            "Float list of z_switch has to have length exactly one less \
+                            than list of layer values"
+        )
+    if len(z_switch) == 0:
+        raise ValueError("Float list of z_switch cannot be empty")
+    for i in range(len(z_switch)):
+        if i == 0:
+            cond = fire.conditional(
```
ndf-ti: i think you can start the loop from one and then set cond outside of this. This way you will not need to check the if in every iteration

spyro/io/interpolators.py
```diff
@@ -0,0 +1,258 @@
...
+        x_index = np.searchsorted(x_nodes, source_coords[:, 1])
+        source.dat.data[:] = vp_values[z_index, x_index]
+    else:
+        z_nodes = np.unique(source_coords[:, 0])
+        x_nodes = np.unique(source_coords[:, 1])
+        y_nodes = np.unique(source_coords[:, 2])
+        z_index = np.searchsorted(z_nodes, source_coords[:, 0])
+        x_index = np.searchsorted(x_nodes, source_coords[:, 1])
+        y_index = np.searchsorted(y_nodes, source_coords[:, 2])
+        source.dat.data[:] = vp_values[z_index, x_index, y_index]
```
ndf-ti, line 52: wdyt of setting the z_nodes and z_index outside of the ifs? Maybe something like this can work:

```Python
 z_nodes = np.unique(source_coords[:, 0])
x_nodes = np.unique(source_coords[:, 1])
z_index = np.searchsorted(z_nodes, source_coords[:, 0])
x_index = np.searchsorted(x_nodes, source_coords[:, 1])

if vp_values.ndim == 2:
    source.dat.data[:] = vp_values[z_index, x_index]
else:
    y_nodes = np.unique(source_coords[:, 2])
    y_index = np.searchsorted(y_nodes, source_coords[:, 2])
        source.dat.data[:] = vp_values[z_index, x_index, y_index]
```

spyro/io/interpolators.py
```diff
@@ -0,0 +1,258 @@
...
+        y_extent = Model.mesh_parameters.length_y + 2.0 * pad_length
+        spacing_z = z_extent / float(vp_values.shape[0] - 1)
+        spacing_x = x_extent / float(vp_values.shape[1] - 1)
+        spacing_y = y_extent / float(vp_values.shape[2] - 1)
+        grid_spacing = (
+            spacing_z
+            if np.isclose(spacing_z, spacing_x) and np.isclose(spacing_z, spacing_y)
+            else None
+        )
+        length_y = Model.mesh_parameters.length_y
```
ndf-ti, line 93: same comment as above, some of the code in the if and else block are repeated

spyro/io/interpolators.py
Olender: This is leftover code from when we used a buffer and the whole mesh might have been slightly dislocated, which doesn't happen anymore. I will remove this line

demos/acoustic_layers_fwi.py
ruansava: Comment from box in "conversation" tab

ruansava: Comment from box in "file changed" tab starting a review

ruansava (commented):
Comment form box "finish you review"

spyro/io/interpolators.py
Olender: I just moved these methods as is from the overcrowded basicio to a new .py file, but that's a good idea. I will change it

Olender: Done

spyro/io/interpolators.py
Olender: ok

spyro/io/interpolators.py
Olender: CHanged to 0.0

demos/advanced_simple_fwi_wrapper.py
Olender: Done!

spyro/io/interpolators.py
Olender: DOne

demos/acoustic_marmousi_forward_propagation.py
```diff
@@ -0,0 +1,86 @@
...
+}
+dictionary["visualization"] = {
+    "forward_output": True,
+    "forward_output_filename": "results/forward_output.pvd",
+}
+
+
+def test_real_shot_record_generation_parallel():
+    real_dictionary = deepcopy(dictionary)
+    real_dictionary["mesh"]["mesh_file"] = "meshes/real5hz.msh"
```
rasjr1305, line 76: Correct me if generating the mesh in parallel would cause problems here, but I would prefer to generate it in real time here instead of importing a .msh, that would allow to change properties and see results quickly

Olender: The marmousi case with mesh generation is a different branch based on this one, and the PR will be opened as soon as this merges into main. I have split up what would be a bigger PR into smaller ones

PR #337, opened 2026-07-28
DO NOT SQUASH: Simplify modal solver (1)

ruansava:
1. Refactoring in the modal solver simplify the use in the code.
2. Creation of class `Modal_Analytical_Solver`.
3. Improving documentation.
4. Creation of new types of error methods to validate data structures.


spyro/solvers/modal/modal_ana_sol.py
```diff
@@ -0,0 +1,748 @@
...
+
+        if type_homog == "energy":
+            # Equivalent velocity by energy-equivalent homogenization
+
+            # Weak forms
+            a, L = weak_forms(c, V, quad_rule=quad_rule, source=True,
+                              user_load=static_load_for_ceq)
+
+            # Compute the energy
+            solve(a == L, u)
```
Olender: We need to check the efficiency of setting up the solver this way

spyro/utils/error_management.py
```diff
@@ -119,12 +123,14 @@ def clean_inst_num(data_arr):
     data_arr : `array`
         An array with null or positive components.
     """
+    type_data_structure_error("data_arr", data_arr, "array",
+                              expected_type_element=("float", "int"))
     data_arr[where(isnan(data_arr) | isinf(data_arr) | (data_arr < 0.0))] = 0.0
     return data_arr
```
Olender, line 129: Could this end up masking a bigger problem in io? Maybe we should add a warning (although there are already a lot of those in the code)

spyro/utils/error_management.py
```diff
@@ -269,8 +284,149 @@ def value_string_error(par_name, par_value):
...
+    expected_valid += (WithGeometry,) if "FunctionSpace" in expected_type else ()
+    if not isinstance(par_value, expected_valid):
+        opt_str = ", ".join([f"'{etype}'" for etype in expected_valid])
+        last_comma = opt_str.rfind(',')
+        opt_str = opt_str[:last_comma] + " or" + opt_str[last_comma + 1:] \
+            if len(expected_valid) > 1 else opt_str
+        raise TypeError(f"'{par_name}' must be of type: {opt_str}, "
+                        f"got {type(par_value).__name__}.")
+
+    return par_value
```
Olender, line 432: In what cases would this be used? To get a clearer error before something goes wrong in a form or assemble?

spyro/utils/error_management.py
```diff
@@ -269,8 +284,149 @@ def value_string_error(par_name, par_value):
...
+    # Check if all elements are of expected type
+    if expected_type_element is not None:
+        if isinstance(expected_type_element, str):
+            expected_type_element = (expected_type_element,)
+        for etype in expected_type_element:
+            value_parameter_error("expected_type_element", etype,
+                                  ["float", "int", "str", "NoneType"])
+        expected_types = tuple(element_map[etype] for etype in expected_type_element)
+        expected_types += (int64,) if "int" in expected_type_element else ()
+        expected_types += (float64,) if "float" in expected_type_element else ()
```
Olender, line 372: When we input segy data, sometimes we can get an implicit float32 conversion to float64, would this end up generating an error in those cases?

tests/on_one_core/test_modal_analytical_solver.py
```diff
@@ -0,0 +1,545 @@
...
+    str_ele = element_geometry + "_" + ("Hom" if homogeneous else "Het")
+    dictionary["visualization"] = {  # Output folder
+        "output_folder": f"output/modal_analytical_test{dimension}d"
+        + f"/modal_analytical_test{dimension}d" + str_ele
+    }
+
+    return dictionary
+
+
+@fixture(scope="function")
```
Olender, line 112: I have to figure out what this does

Olender (changes requested):
I think this is a fairly straightforward PR, since most of the changes consist of moving and reorganizing the existing code from the modal solver into a cleaner structure. I like the refactoring overall.

My only concern is the use of the Firedrake solver object with implicit/default solver parameters. Since solve(a==L) hides the underlying PETSc configuration, it isn't immediately clear which solver and preconditioner are actually being used, or which ones this implementation expects. Would it be possible to explicitly specify the solver parameters (or at least document the expected defaults)? I think that would improve reproducibility and make the solver behavior more transparent, while also safeguarding ourselves against changing defaults in FIredrake or PETSc which might cause some really difficult debugging in the future.

spyro/utils/error_management.py
ruansava: Depends on what you want it for. I use it only to clean the initial guess for the non-linear eikonal

spyro/solvers/modal/modal_ana_sol.py
ruansava: Ok

spyro/utils/error_management.py
ruansava: We can add this check to expected types

spyro/utils/error_management.py
ruansava: I use it in public methods where I don't employ class attributes. That is, I have to pass an argument to the function. See `spyro.solvers.modal,modal_sol` or `spyro.solvers.modal,modal_ana_sol`

tests/on_one_core/test_modal_analytical_solver.py
ruansava: I added support for int32 and float32

spyro/solvers/modal/modal_ana_sol.py
ruansava: It is a simple static dummy problem. I don't know if we should use the linear solver explicitly.

spyro/utils/error_management.py
ruansava: For the case `type_data_structure_error` method, we can only check if the data structure corresponds to the expected type, without checking the type of elements it contains.

spyro/solvers/modal/modal_ana_sol.py
ruansava: I am going to use the solver setting used in PML

ruansava: Linear solver implemented!

spyro/solvers/modal/modal_ana_sol.py
```diff
@@ -0,0 +1,766 @@
...
+            # Weak forms
+            a, L = weak_forms(c, V, quad_rule=quad_rule, source=True,
+                              user_load=static_load_for_ceq)
+
+            # Solve static load problem for the energy-equivalent homogenization
+            lin_var = LinearVariationalProblem(a, L, u, constant_jacobian=True)
+            fem_method = "spectral_quadrilateral" if self.quadrilateral \
+                else "mass_lumped_triangle"
+            solver_parameters = get_default_parameters_for_method(fem_method)
+            solver_parameters["mat_type"] = "matfree"
```
Olender:
```suggestion
            solver_parameters = {
                "mat_type": "matfree",
                "ksp_type": "cg",
                "pc_type": "gamg",
            }
```

ruansava: It did not work :(

ruansava: new parameter set that works
            solver_param = {"ksp_type": "gmres",
                            "pc_type": "hypre",
                            "pc_hypre_type": "boomeramg",
                            "ksp_rtol": 1e-12,
                            "ksp_atol": 1e-12,
                            "ksp_gmres_restart": 100}

Olender (approved):

PR #338, opened 2026-07-28
DO NOT SQUASH: Test error manager (2)

ruansava:
1. Test suite of the methods implemented in `spyro.utils.error_management`
2. Fixing some bugs related to the error methods

### DIFF
--- a/spyro/abc/abc_layer.py
+++ b/spyro/abc/abc_layer.py
@@ -227,7 +227,7 @@ class ABCLayer(NRBC):
         # Nyquist frequency
         self.freq_Nyquist = value_numerical_error("freq_Nyquist", freq_Nyquist,
                                                   float_num=True, integer_num=True,
-                                                  lower_bound=0.)
+                                                  lower_bound=0., none_default=True)
 
         # Model dimension
         self.dimension = value_parameter_error("dimension", dimension, [2, 3])
--- a/spyro/abc/hyp_lay.py
+++ b/spyro/abc/hyp_lay.py
@@ -454,7 +454,8 @@ class HyperLayer():
                               integer_num=True, lower_bound=0.)
 
         # Checking the minimum mesh size
-        value_numerical_error('lmin', lmin, float_num=True, lower_bound=0.)
+        value_numerical_error('lmin', lmin, float_num=True,
+                              integer_num=True, lower_bound=0.)
 
         # Domain dimensions
         length_z, length_x = self.domain_dim[:2]
--- a/spyro/utils/error_management.py
+++ b/spyro/utils/error_management.py
@@ -3,7 +3,7 @@
 This file contains methods for handling errors in Spyro, either to send
 messages to the user or to prevent numerical instability in objects."""
 
-from numpy import float64, inf, int64, isinf, isnan, ndarray, where
+from numpy import float64, inf, int64, isinf, isnan, ndarray
 from firedrake import Function, FunctionSpace, Mesh
 from firedrake.functionspaceimpl import WithGeometry
 from ufl.geometry import SpatialCoordinate
--- a/spyro/utils/error_management.py
+++ b/spyro/utils/error_management.py
@@ -66,17 +66,19 @@ def mutually_exclusive_parameter_error(par_name_lst, par_value_lst):
                                            par_value_lst)
                    if val is not None]
 
-    # Error message about the invalid parameter
-    exc_str = "Parameters " + ", ".join([f"'{name}'" for name in par_defined])
-    last_comma = exc_str.rfind(',')
-    exc_str = exc_str[:last_comma] + ' and' + exc_str[last_comma + 1:]
-    exc_str += " mutually exclusive.\n"
-    err_str = "Please specify only one of these parameters: "
-    opt_str = ", ".join([f"'{val}'" for val in par_name_lst])
-    last_comma = opt_str.rfind(',')
-    opt_str = opt_str[:last_comma] + " or" + opt_str[last_comma + 1:]
+    # Only raise if two or more parameters are defined
+    if len(par_defined) > 1:
+        # Error message about the invalid parameter
+        exc_str = "Parameters " + ", ".join([f"'{name}'" for name in par_defined])
+        last_comma = exc_str.rfind(',')
+        exc_str = exc_str[:last_comma] + ' and' + exc_str[last_comma + 1:]
+        exc_str += " mutually exclusive.\n"
+        err_str = "Please specify only one of these parameters: "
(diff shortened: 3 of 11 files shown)


spyro/abc/abc_layer.py
```diff
@@ -250,18 +273,10 @@ def __init__(self, domain_dim, frequency, freq_Nyquist, dimension=2,
...
-
-        # # Initializing the error measure class
-        # HABCError.__init__(self, self.dt, self.freq_Nyquist,
-        #                    self.receiver_locations,
-        #                    output_folder=self.path_save,
-        #                    output_case=self.path_case_abc)
+        # Initializing the error measure class
+        # HABCError.__init__(
+        #     self, self.dt, freq_Nyquist, self.receiver_locations,
+        #     output_folder=self.path_save, output_case=self.path_case_abc, comm)
```
Olender, line 279: These lines can be removed now right?

spyro/abc/abc_layer.py
```diff
@@ -782,12 +797,12 @@ def velocity_abc(self, Wave, inf_model=False, method="point_cloud", save_file=Tr
...
             outfile.write(Wave.c)
 
-    def nrbc_on_boundary_layer(self, Wave_object, non_reflect_bc, save_file=True):
+    def nrbc_on_boundary_layer(self, Wave, non_reflect_bc, save_file=True):
         """Apply Non-Reflective BCs on the outer boundary of the absorbing layer.
 
         Parameters
         ----------
-        Wave_object : `acoustic_wave.AcousticWave`
+        Wave : `acoustic_wave.AcousticWave`
```
Olender, line 805: why the cahnge? wouldnt a better change be from Wave_object to wave_object?

Olender: Now it is the same name as the Wave Class which is a bit confusing

spyro/abc/hyp_lay.py
```diff
@@ -150,7 +150,7 @@ def define_hyperaxes(self, domain_hyp):
         domain_hyp : `tuple`
             Domain dimensions with layer without truncation by free surface.
             2D : (length_z + 2 * pad_len, length_x + 2 * pad_len).
-            3D : (length_x + 2 * pad_len, length_x + 2 * pad_len, length_y + 2 * pad_len).
+            3D : (length_z + 2 * pad_len, length_x + 2 * pad_len, length_y + 2 * pad_len).
```
Olender, line 153: SHould this really be x2 pad for length z?

spyro/abc/abc_layer.py
ruansava: I follow this: All things in python are objects, so that every argument here is one of these things. Why should we specify that it is an object? A simple example, if I use: `names_of_someone = list(['Ruben', 'Andres', 'Alexandre'])`, why would I use `names_of_someone_object = list(['Ruben', 'Andres', 'Alexandre'])`?

spyro/abc/abc_layer.py
ruansava: No yet! :v

spyro/abc/hyp_lay.py
ruansava: For the calculation of the hyper-axes is necessary. For this reason, it is required "layer without truncation"

spyro/abc/abc_layer.py
ruansava: `Wave` is an abstract class, then to instance is not allowed.
I believe names should be assigned throughout the code based on the phenomenon, such as wave_acu, wave_ela, wave_ani, or something like that. In the future, the question is not whether it is an object, but rather what type of problem I am modeling.

Olender: My concern in slightly different now. Not about using `wave` versus `wave_object`. It is about using `Wave` (with a captal letter) as the variable name. 

I would like to thank you for removing the `_object`, this is very valid, as you have pointed out

Olender: According to **PEP 8**, class names should use `CapWords` (e.g. `Wave`), while variables and instances should use `lower_case_with_underscores` (e.g. `wave`, `wave_acoustic`). Since `Wave` is the name of the abstract class, using `Wave` as the name of an instantiated object is confusing because it blurs the distinction between the type and the instance

I will also add a mea-culpa. All throughout spyro I've used `Wave_object` because I did not know, at the time, about those naming conventions and wanted to differentiate between the instantiated object and the class name.

Olender: For that reason, I think names such as `wave`, `wave_acoustic`, `wave_elastic`, or `wave_anisotropic` are clearer. They describe the modeled phenomenon while also following the usual Python convention that classes are `CapWords` and instances are lowercase

spyro/abc/hyp_lay.py
Olender: Okay, thanks for the clarification

tests/on_one_core/test_error_manager.py
Olender, line 1: I've never used tests like these (with a `Class`, whose name starts with `Test` and methods with `test` in their name inside). How do these work (I'm not saying I disapprove, I'm just curious on their inner workings with pytest and why they were chosen instead of methods)?

Olender: This looks very useful for reorganizing some tests in the future (specifically the io based ones that use a lot of the same variables across different methods)

spyro/utils/error_management.py
```diff
@@ -110,26 +112,55 @@ def value_model_dimension_error(par_names, parameters, expected_dim):
         raise ValueError(dim_err)
 
 
-def clean_inst_num(data_arr):
-    """Set NaNs and negative values to zero in an array.
+def clean_inst_num(data_arr, nan_values=True, inf_values=True, negative_values=True):
```
Olender, line 115:
```suggestion
def sanitize_numeric_array(data_arr, nan_values=True, inf_values=True, negative_values=True):
```

I find this name a bit criptic, what do you think about this change?

Olender (approved):

spyro/utils/error_management.py
ruansava: Nicce! I will change it in next PR.

tests/on_one_core/test_error_manager.py
ruansava: I saw the structure and liked it, as it allows for better organization and grouping of elements, as well as sharing configuration code across tests.

spyro/abc/abc_layer.py
ruansava: Ok!

PR #341, opened 2026-07-31
Correct wave object name

Olender:
TLDR: Let's follow PEP8 and change Wave_obj and similar capitalized Wave instantiated based objects to wave

This is to fix an error I made in the past that spread to the complete codebase


spyro/plots/plots_habc.py
```diff
@@ -173,7 +173,7 @@ def plot_function_layer_size(lay_par, freq_par, geom_par, FLpos,
 
 #     Parameters
 #     ----------
-#     Wave_object : `wave`
+#     wave : `wave`
```
ruansava, line 176: Please modify where needed
        Wave : `acoustic_wave.AcousticWave`
            An instance of the :class:`~spyro.solvers.acoustic_wave.AcousticWave`.

ruansava (approved):

spyro/plots/plots_habc.py
Olender: Just to make sure I understand:

Are you suggesting that the parameter name wave should be renamed while keeping the type as acoustic_wave.AcousticWave, or is there another change you have in mind?

Olender: I think I understood what you meant and did it in the latest commit

PR #355, opened 2026-08-06
Append snapshot during the wave time integration

Ig-dolci:
Snapshots are appended as the solve advances rather than allocated up front. The final footprint is the same, but nothing is reserved before it is needed: a solve that aborts early (e.g. the numerical instability check below) no longer has to first allocate the whole nt-step wavefield, which for fine meshes/small dt is many GB.

### DIFF
--- a/spyro/solvers/time_integration_central_difference.py
+++ b/spyro/solvers/time_integration_central_difference.py
@@ -36,11 +36,12 @@ def _propagate_forward_central_difference(wave, source_ids):
     nt = int(wave.final_time / wave.dt) + 1  # number of timesteps
     usol = None
     if wave.store_forward_time_steps:
-        usol = [
-            fire.Function(wave.function_space, name=wave.get_function_name())
-            for t in range(nt)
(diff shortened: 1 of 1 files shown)


Olender (changes requested):
This 'request changes' is just to stop any merge to main (github is partially down, they even posted in https://www.githubstatus.com/) and its doing weird things to our runners and backup tag generation

spyro/solvers/time_integration_central_difference.py
```diff
@@ -67,6 +68,10 @@ def _propagate_forward_central_difference(wave, source_ids):
     usol_recv = []
     receiver_array = None
     receiver_buffer = None
+    # Reused accumulator for the point-source cofunction, so the per-timestep
+    # source assembly writes into an existing tensor instead of allocating a
+    # fresh Cofunction on every step.
+    source_buffer = None
```
Olender, line 74: This is interesting

Olender (approved):
Just don't merge yet, since github actions is still mostly down

PR #358, opened 2026-08-07
Refactor old plots and add new elastic wave and comparison plots

Olender:
TLDR: Split `plots.py` into modules, added reusable plotting utilities, displacement plots for elastic, and converted commented-out code into maintainable functionality.

### Description

The `plots.py` file was large and hard to mantain. Also, @Ig-dolci on my previous analytical elastic PR, asked to separate the elastic displacement plotting functionality with the analytical elastic solution implementation.

As part of this cleanup, I also refactored one of @ruansava's large commented-out code sections into a proper receiver comparison plotting function. Since Git already preserves the project's history, converting commented code into reusable functionality helps keep the codebase cleaner and easier to maintain.

This is a draft PR since I am adding notebooks to show the new plotting functionality and some tests might still need adjusting.

------------------------------
Disclaimer: since its 4h31 pm on a Friday, the text section written below this one uses copilot to format and analyze
------------------------------

### Main changes

* Split the monolithic `plots.py` into focused modules:

  * `general_plots.py`
  * `mesh_plots.py`
  * `debug_plots.py`
  * `receiver_plots.py`
* Added a shared plotting helper (`plot_helpers.py`) to centralize figure saving and display logic.
* Added dedicated plotting routines for:

  * mesh visualization,
  * debugging Firedrake functions,
  * receiver time-history comparisons,
  * displacement component visualization.
* Updated imports throughout the codebase to use the new module structure.
* Added and updated tests for the new plotting functionality.

### Benefits

* Smaller, more maintainable plotting modules.
* Clearer separation of responsibilities.
* Reusable plotting utilities instead of commented-out code.
* Improved documentation and test coverage.
* Easier future development of plotting features.


spyro/plots/mesh_plots.py
```diff
@@ -0,0 +1,103 @@
...
+    restores them afterwards to avoid side effects.
+    """
+    plt.rcParams["font.size"] = 12
+
+    if mesh_filename is not None:
+        mesh = Mesh(mesh_filename)
+    elif firedrake_mesh is not None:
+        mesh = firedrake_mesh
+    else:
+        raise ValueError("Please specify mesh")
```
ndf-ti: why accept a mesh_filename? i know it can be convenient but having lots of optional parameters can make the code harder to read. WDYT of making this function only accept the firedrake_mesh arg and flag it as required?

spyro/plots/mesh_plots.py
```diff
@@ -0,0 +1,103 @@
...
+        cbar = fig.colorbar(im, orientation="horizontal")
+        cbar.ax.set_xlabel("circumcircle radius (km)")
+    fig.set_size_inches(13, 10)
+    _finalize_figure(fig, filename=output_filename, show=show)
+
+    # Flip back mesh coordinates so it does not change outside of method
+    coordinates = mesh.coordinates.dat.data.copy()
+
+    mesh.coordinates.dat.data[:, 0] = coordinates[:, 1]
+    mesh.coordinates.dat.data[:, 1] = coordinates[:, 0]
```
ndf-ti, line 95: question: one of the method can change the data inside the mesh so you need to undo the changes here? is it possible to copy the mesh itself to do this so you dont have to make the undo by yourself?

spyro/plots/plot_helpers.py
```diff
@@ -0,0 +1,33 @@
...
+            # Save exactly what the caller requested.
+            fig.savefig(filename, **savefig_kwargs)
+        else:
+            # Replace or add extensions.
+            stem = filename.with_suffix("")
+            for fmt in formats:
+                fig.savefig(
+                    stem.with_suffix(f".{fmt}"),
+                    **savefig_kwargs,
+                )
```
ndf-ti, line 28: question: do we really need to deal with formats? if that is not a hard requirement, i think we should just drop the format handling

spyro/plots/receiver_plots.py
```diff
@@ -0,0 +1,366 @@
...
+    if name is not None:
+        line_kwargs.setdefault("label", name)
+
+    ax.plot(time_vector, receiver_data, **line_kwargs)
+    ax.set_xlabel("time (s)", fontsize=18)
+    ax.set_ylabel("receiver response", fontsize=18)
+    if receiver_id_for_title is not None:
+        ax.set_title(f"Receiver ID {receiver_id_for_title} data.", fontsize=22)
+    else:
+        ax.set_title("Receiver data.", fontsize=22)
```
ndf-ti, line 92: wdyt of making the title a required argument?

spyro/plots/receiver_plots.py
```diff
@@ -0,0 +1,366 @@
...
+def plot_displacement_components(
+    time_vector: np.ndarray | list[float] | tuple[float, ...],
+    displacement_tuple: tuple[
+        np.ndarray | list[float] | tuple[float, ...],
+        np.ndarray | list[float] | tuple[float, ...],
+        np.ndarray | list[float] | tuple[float, ...],
+    ],
+    source_type: str = "Unknown",
+    save_plots: bool = False,
+    output_dir: str | Path | None = None,
```
ndf-ti: i think we can just keep output_dir. If the variable is set, we assume that the user wants to save the plot

spyro/plots/mesh_plots.py
Olender: YOu're right. It's unnecessary. The user can pass a mesh that's already loaded by Firedrake

Olender: Done

spyro/plots/plot_helpers.py
Olender: I don't like them, but another user always uses them and I don't want to drop support for him

spyro/plots/receiver_plots.py
Olender: Because I'm lazy and don't want to type a title, specially since most titles are going to be the same

spyro/plots/receiver_plots.py
Olender: Makes sense. I'm changing it

spyro/plots/mesh_plots.py
Olender: I couldn't figure out how to copy the mesh object. I couldn't use the copy library for it. However, I would prefer your way. It doesn't really make sense to have to modify the data inside the mesh object and than change it back again

Olender:
The tests are adjusted. This PR is a draft because I still would like to add some notebooks exemplifying plot functionality

ruansava (approved):

ndf-ti (approved):

PR #359, opened 2026-08-07
Added 3D meshing with Gmsh

rasjr1305:
I added now the 3D meshing functionality to Spyro, the code is big because I needed to create the 3D rectangular padding step by step, for it to work with structured meshes. The following meshes should work:

Unstructured
No water, rectangular or elliptical padding, tetrahedrals
Water Interface, rectangular or elliptical padding, tetrahedrals

Structured
No water, rectangular padding, hexahedrals
Water Interface, rectangular padding, hexahedrals

While the PR is availed I will be adding features like tutorials and the corrections needed. I will also try to get the structured code smaller if possible.

- Added a in-house free velocity model to be used in examples;
- Added the 2D meshing tutorial;
- Added the 3D meshing tutorial;


spyro/meshing/meshing_functions.py
```diff
@@ -539,7 +550,7 @@ def create_gmsh_2D_mesh(self):
                 domain_xmax = box_xmax + padding_x
 
             # Interpolating sizing function from segy file
-            ef_segy2, f_min, f_max, n_samples, n_traces = create_sizing_function(
+            ef_segy2, f_min, f_max, n_samples, n_traces = create_sizing_function(  # noqa: RUF059
```
Olender: RUF059 is an unused variable right? Usually we just substitute it for `_`, for example if `y` was unusned in 
```
x, y = something(), 
```
we can change it to:
```
 x, _ = something()
```

spyro/meshing/meshing_parameters.py
```diff
@@ -65,7 +65,7 @@ def cells_per_wavelength(method, degree, dimension):
     return cell_per_wavelength_dictionary.get(key)
 
 
-class MeshingParameters():
+class MeshingParameters():  # noqa: UP039
```
Olender, line 68: I like this. You are in the future with all these ruff error avoids. Yes, I am thinking of switching to a mix of black than ruff (with check and fix options) in the very near future. Although, I'll probably just add an ignore to this rule.

spyro/meshing/meshing_winslow3D.py
```diff
@@ -0,0 +1,901 @@
...
+        )
+        z_bottom = (
+            z_coordinates[bottom]
+            if bottom != -1
+            else (
+                2.0 * z_center - z_coordinates[top]
+                if top != -1
+                else z_center
+            )
+        )
```
ndf-ti: i think you can extract this to a function, like:

```python
def ...(coordinates, dir_index, neg_dir_index, center):
     return coordinates[dir_index]
              if dir_index != -1
              else (
                  2.0* center - coordinates[neg_dir_index]
                  if neg_dir_index!= -1
                  else center
              )
```

spyro/meshing/meshing_winslow3D.py
```diff
@@ -0,0 +1,901 @@
...
+        x_nw = (
+            x_coordinates[north_west]
+            if north_west != -1
+            else x_north + x_west - x_center
+        )
+        x_se = (
+            x_coordinates[south_east]
+            if south_east != -1
+            else x_south + x_east - x_center
+        )
```
ndf-ti: same thing as above, you cant extract this logic to a function

spyro/meshing/meshing_functions.py
rasjr1305: I am using the Ruff package to automaticaly correct the Lint errors, it seems I leaved that comment for Ruff ignore that line

spyro/meshing/meshing_winslow3D.py
rasjr1305: I will take a look into that

spyro/meshing/meshing_winslow3D.py
rasjr1305: thanks i will simplify this

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+    4. no padding, water delimitation;
+    5. rectangular padding, water delimitation;
+    6. hyperelliptical padding, water delimitation.
+
+    Returns
+    -------
+    dict
+        Geometry and volume tags needed by mesh generation and smoothing.
+    """
+    del minElementSize
```
Olender: WHy the del?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+    nx,
+    ny,
+    dz,
+    dx,
+    dy,
+    byte_order="big",
+    axes_order=(0, 1, 2),
+    axes_order_sort="F",
+    dtype="float32",
+):
```
ndf-ti, line 42: wdyt of encapsulating related arguments together here? like, putting dx, dz, dy in one object? You can event create a data class to hold this values like:

```python
@dataclass
class Vector:
     x: float
     y: float
     z: float
````

then you can access it like:

```Python
length = Vector(1, 1, 1)

lenght.x
```

spyro/meshing/meshing_gmsh3D.py
ndf-ti: is this needed?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            "in (z, x, y) order; axes_order must therefore be (0, 1, 2)."
+        )
+
+    domainX = float(length_x)
+    domainY = float(length_y)
+    domainZ = abs(float(depth_z))
+    padX = float(padding_x)
+    padY = float(padding_y)
+    padZ = float(padding_z)
+    ellipse_n = float(hyper_n)
```
ndf-ti: instead of converting like this, you can use type hints in the function arguments

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+
+    ORDER = axes_order_sort
+    DTYPE_BE = np.dtype(dtype).newbyteorder(">" if byte_order == "big" else "<").str
+
+    ellipseLx = padX
+    ellipseLy = padY
+    ellipseLz = padZ
```
ndf-ti: is this needed? i think you can use padding_x instead

spyro/meshing/meshing_gmsh3D.py
ndf-ti: you can event add a validation in dataclass to assert that all the fields are bigger than 0, so you dont have to do these validations inside the function.

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            if lx <= 0.0 or ly <= 0.0 or lz <= 0.0:
+                return None
+            return occ.addBox(
+                float(x0),
+                float(y0),
+                float(z0),
+                float(lx),
+                float(ly),
+                float(lz),
+            )
```
ndf-ti: i think there is no need to declare the function here. Maybe you can move it to another file, wdyt?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+
+    q = np.asarray(q, dtype=float)
+    print(
+        f"[quality] count={q.size}  min={q.min():.6g}  p1={np.percentile(q, 1):.6g}  "
+        f"p5={np.percentile(q, 5):.6g}  median={np.median(q):.6g}  "
+        f"p95={np.percentile(q, 95):.6g}  max={q.max():.6g}  mean={q.mean():.6g}"
+    )
+
+
+def create_super_ellipsoid_volume(
```
ruansava: Replace superellipsoid with hyperellipsoid wherever necessary, since n >= 2 to ensure a convex volume that surrounds the original domain

ruansava (changes requested):

spyro/meshing/meshing_functions.py
Olender: Instead of ignoring the error you can just use the underline

notebook_tutorials/meshing_2d.ipynb
Olender, line 1: Yes! New tutorials!

spyro/meshing/meshing_functions.py
Olender: Here you can do this to f_min and f_max

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
+from pathlib import Path
+
```
Olender: Why the space between the pathlib and the numpy module import?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+import numpy as np
+
+from .meshing_winslow3D import winslow_smooth_3d55
+
+
+def report_quality(gmsh, dim=3, quality_type=2):
```
Olender: I always annoy @ruansava with this as well, but, why the short variable names? You can write dimension here. Quoting from PEP20, the zen of python: "Readability counts".

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+    xc, yc, zc : float
+        Center coordinates of the ellipsoid
+
+    Returns:
+    --------
+    volume_tag : int
+        OpenCASCADE volume tag
+    """
+
+    def super_ellipsoid_point(u, v, a, b, c, n, xc, yc, zc):
```
Olender: Can't this method be outside of the other method?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            print("Successfully created volume using closed B-spline surface")
+    except Exception as exc:
+        raise RuntimeError(
+            "B-spline surface creation failed."
+        ) from exc
+
+    return volume_tag
+
+
+def create_properly_closed_surface(gmsh, a, b, c, n, xc, yc, zc, point_func):
```
Olender: WHat is point_func? Please add more docs

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+    """
+    Create a closed B-spline surface by ensuring u-direction closure and including poles.
+    """
+    u_res = 60  # Points in longitude
```
Olender: what about u_resolution?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+
+        # Create volume
+        surface_loop = gmsh.model.occ.addSurfaceLoop([surface_tag])
+        volume_tag = gmsh.model.occ.addVolume([surface_loop])
+        gmsh.model.occ.synchronize()
+
+        # Z cut
+        z_cut = 0.0
+        if volume_tag is not None and z_cut is not None:
+            print(f"Applying z-cut at z = {z_cut}")
```
Olender: remember to always flush prints if you are not using spyro's parallel_print method

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+
+    volume_tag = None
+
+    try:
+        volume_tag = create_properly_closed_surface(
+            gmsh, a, b, c, n, xc, yc, zc, super_ellipsoid_point
+        )
+        if volume_tag:
+            print("Successfully created volume using closed B-spline surface")
+    except Exception as exc:
```
Olender: WHy have this raise an error here instead of just letting it raise an error and fail internally?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            if result[0]:
+                volume_tag = result[0][0][1]
+                print("Z-cut applied successfully")
+            else:
+                print("Warning: Z-cut removed entire volume")
+                volume_tag = None
+
+        return volume_tag
+
+    except Exception as exc:
```
Olender: Same as my other comment

spyro/meshing/meshing_gmsh3D.py
Olender: minElementSize isn't global. Please just remove this

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            "for this geometry."
+        )
+
+    if water_interface and tuple(axes_order) != (0, 1, 2):
+        raise ValueError(
+            "The supplied water-interface geometry reads the binary directly "
+            "in (z, x, y) order; axes_order must therefore be (0, 1, 2)."
+        )
+
+    domainX = float(length_x)
```
Olender: Why the conversions?

spyro/meshing/meshing_gmsh3D.py
Olender: That's a good idea!

spyro/meshing/meshing_gmsh3D.py
Olender: exactly

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+    yc = domainY / 2.0
+    zc = -domainZ / 2.0
+
+    z_min, z_max = float(depth_z), 0.0
+    x_min, x_max = 0.0, float(length_x)
+    y_min, y_max = 0.0, float(length_y)
+
+    if not water_interface and padding_type is None:
+        # No padding
+        print(
```
Olender: Another print that isn't beign flushed. This will cause the print to appear out of order (not when it should) in clusters like mintrop or senai-cimatec. In mintrop it will show up next (right before or after) the next flushed print, in senai's cluster a nonflushed print will just show up randomly in the output (I have no idea why)

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            "Generating undelimited water+subsurface rectangular domain (no padding)..."
+        )
+        occ = gmsh.model.occ
+        vol_subsoil = occ.addBox(
```
Olender, line 155: A descriptive name here would be nice

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            x_min,
+            y_min,
+            z_min,
+            x_max - x_min,
+            y_max - y_min,
+            z_max - z_min,
+        )
+        occ.synchronize()
+
+        gmsh.model.addPhysicalGroup(3, [vol_subsoil], name="SubsurfaceAndWater")
```
Olender, line 165: Ahhh, I love these Physical gropus for gradient calculation. Perfect!

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+        vol_water = None
+
+    elif not water_interface and padding_type == "rectangular":
+        # Rectangular padding
+        print(
```
Olender: Another print

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            "Generating undelimited water+subsurface domain with rectangular padding..."
+        )
+        occ = gmsh.model.occ
+
+        if min(domainX, domainY, domainZ) <= 0.0:
+            raise ValueError("length_x, length_y, and depth_z must be positive.")
```
Olender: Don't you abs depth_z previously? Why is this here for domainZ?

spyro/meshing/meshing_gmsh3D.py
Olender: Adding to @ndf-ti's comment, this method can be outside, typechecked and documented. WIth these non descriptive variable names I've already forgotten who x0 and lx are

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            )
+
+        vol_subsoil = core_fragments[0]
+        vol_water = None
+
+        # both water and subsurface.
+        gmsh.model.addPhysicalGroup(3, core_fragments, name="SubsurfaceAndWater")
+        gmsh.model.addPhysicalGroup(3, padding_fragments, name="Padding")
+
+        def boundary_faces_of_volume(vol_tag):
```
Olender: These functions could go into another file. Like gmsh_helpers.py or something

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,4919 @@
...
+            boundary = gmsh.model.getBoundary(
+                [(3, vol_tag)],
+                oriented=False,
+                recursive=False,
+                combined=True,
+            )
+            return [tag for dim, tag in boundary if dim == 2]
+
+        ellipsoid_faces = set()
```
Olender, line 249: Parei aqui

spyro/meshing/meshing_parameters.py
Olender: We already have black and ruff formating in spyro. just run make format in the terminal!

spyro/meshing/meshing_gmsh3D.py
Olender: The python garbage collector is really good at what it does. Also del does not make the garbage collector collect. It will collect when it wants to

notebook_tutorials/meshing_2d.ipynb
Olender: Thanks for doing this!

spyro/meshing/meshing_gmsh3D.py
rasjr1305: I am fixing and applying the corrections

spyro/meshing/meshing_gmsh3D.py
ruansava: hyper_ellipsoid_point(u, v, a, b, c, n, xc, yc, zc):

spyro/meshing/meshing_gmsh3D.py
ruansava: > remember to always flush prints if you are not using spyro's parallel_print method

Import parallel_print from spyro.io.basicio
Thus, you can use as parallel_print(string) without be worried with flush flag

spyro/meshing/meshing_gmsh3D.py
ruansava: If I understood well, pad could be different but hyperellipse won't centered at the centroid of the original domain

spyro/meshing/meshing_gmsh3D.py
ruansava: Use parallel_print as commented above

spyro/meshing/meshing_gmsh3D.py
ruansava: @Olender Question: When we have groups, is the gradient computed at the group boundary as if the domain were continuous? or is there some computation due to the "ficticious" boundary?

spyro/meshing/meshing_winslow3D.py
ruansava: Implement something as (Adjust the variable names if my interpretation isn't correct):
```
def sector_coordinates(coord_matching_direction, coord_reverse_direction, direction):
     coord_sector = coord_matching_direction[direction] if direction != -1  else coord_reverse_direction
```
Use:

`x_sw = sector_coordinates(x_coordinates, x_south + x_west - x_center, south_west)`

Perform a loop for this

spyro/meshing/meshing_winslow3D.py
```diff
@@ -0,0 +1,901 @@
...
+                    + coefficient_f * x_xi_zeta
+                )
+                - source_x
+            ) / denominator
+            x_new[node] = (
+                (1.0 - omega) * x_center
+                + omega * x_target
+            )
+
+        if move_y[node]:
```
ruansava, line 436: There's a lot of code that can be refactored to avoid code repetition

ruansava (changes requested):

Olender:
> @Olender Question: When we have groups, is the gradient computed at the group boundary as if the domain were continuous? or is there some computation due to the "ficticious" boundary?

@ruansava, I can answer for the acoustic case. The gradient is essentially of the form

$$
\sum \int_0^T \int_\Omega (\cdots),\frac{\partial^2 \lambda}{\partial t^2} u_i dx,dt.
$$

When groups are used, I restrict ($\Omega$) and (dx) to the corresponding group/tag. So the integral is computed independently over every tagged volume, including its boundaries. There isn't an additional computation treating the group boundary as a physical/fictitious boundary; it is simply the boundary of the subdomain over which that part of the integral is evaluated.

spyro/meshing/meshing_gmsh3D.py
Olender: @ruansava, I can answer for the acoustic case. The gradient is essentially of the form

$$
\sum \int_0^T \int_\Omega (\cdots),\frac{\partial^2 \lambda}{\partial t^2} u_i dx,dt.
$$

When groups are used, I restrict ($\Omega$) and (dx) to the corresponding group/tag. So the integral is computed independently over every tagged volume, including its boundaries. There isn't an additional computation treating the group boundary as a physical/fictitious boundary; it is simply the boundary of the subdomain over which that part of the integral is evaluated.

spyro/meshing/meshing_winslow3D.py
rasjr1305: Created a function to reduce the size of the code

spyro/meshing/meshing_winslow3D.py
rasjr1305: Done !

spyro/meshing/meshing_gmsh3D.py
rasjr1305: Solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: Will study encapsulation in the end of the pr analysis

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: this is strangely being asked by ruff, we may just ignore

spyro/meshing/meshing_gmsh3D.py
rasjr1305: will take a look at this

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: will add the explanations

spyro/meshing/meshing_gmsh3D.py
rasjr1305: The B-spline is parametrized in the u-v directions along a hyperellipse, U and V are the number of points used to interpolate the spline

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: will take a look at this

spyro/meshing/meshing_gmsh3D.py
rasjr1305: same as above

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: Ok ! Will be adding better variables names.

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_gmsh3D.py
rasjr1305: solved

spyro/meshing/meshing_winslow3D.py
rasjr1305: will take a look at that

spyro/meshing/meshing_winslow3D.py
```diff
@@ -0,0 +1,743 @@
+import sys
+
+import numpy as np
+from numba import njit, prange
```
Olender: You already have the code working both with and without Numba. Please make Numba a soft dependency, similar to what we have done to Gmsh and SeismicMesh.

You can use a `try/except ImportError` block like the ones we have previously used for Gmsh and SeismicMesh. The only difference here is that, since you use decorators imported from Numba, in the `except ImportError` block you would need to define a dummy decorator, something along the lines of:

```python
def njit(func):
    return func
```

This way, the code can still run without Numba installed, while taking advantage of Numba when it is available

Olender (changes requested):

spyro/meshing/meshing_winslow3D.py
rasjr1305: _Yes I forgot that I needed to do that for Numba_

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -687,9 +546,9 @@ def mesh_size_callback3D(dim, tag, x, y, z, lc):
                         return float(graded_size)
 
             gmsh.model.mesh.setSizeCallback(mesh_size_callback3D)
-            print_environment_report()
 
         if parallel:
+            num_threads = os.cpu_count() or 1
```
Olender: Let's not use `os.cpu_count()` here. We should focus on using `mpi4py` for determining the number of MPI processes. You could use something like:

```python
from mpi4py import MPI

num_processes = MPI.COMM_WORLD.Get_size()
```

Or, even better, use the communicator from the `wave` object if it is already available.

In general, we should avoid using `os.cpu_count()` to determine the parallelism of the application. This can cause serious issues when running on an HPC cluster (like mintrop, which is where we run the 3D cases), since the number of CPUs on a node does not necessarily correspond to the number of cores we actually want to use.

For example, suppose I'm running a 3D Overthrust FWI inversion on Mintrop using 5 nodes, each with 40 cores and 192 GB of RAM. I might intentionally use only 10 cores per node because of memory and bandwidth limitations (therefore 50 cores will be available instead of 200). Using `os.cpu_count()` could then result in using more resources than intended, potentially causing severe oversubscription and performance problems.

Also, since we're using Firedrake/PETSc with MPI, I think we should rely on the MPI communicator

Olender (changes requested):

spyro/meshing/meshing_functions.py
```diff
@@ -539,7 +554,7 @@ def create_gmsh_2D_mesh(self):
...
                 fname=fname, hmin=hmin_segy, bbox=segy_bbox, wl=wl, freq=freq,
```
Olender:
```suggestion
            ef_segy2, _, _, n_samples, n_traces = create_sizing_function(
                fname=fname, hmin=hmin_segy, bbox=segy_bbox, wl=wl, freq=freq,
```

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+    depth_z,
+    padding_type,
+    padding_x,
+    padding_y,
+    padding_z,
+    hyper_n,
+    water_interface,
+    water_search_value,
+    structured_mesh,
+    minElementSize,
```
Olender: minElementSize is not accessed. Why do we have to pass it?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+        raise ValueError(
+            "Hyperelliptical 3-D padding currently supports only "
+            "structured_mesh=False. Winslow smoothing is not available "
+            "for this geometry."
+        )
+
+    if water_interface and tuple(axes_order) != (0, 1, 2):
+        raise ValueError(
+            "The supplied water-interface geometry reads the binary directly "
+            "in (z, x, y) order; axes_order must therefore be (0, 1, 2)."
```
Olender, line 128: What if the binary is in (z, x, y)?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+
+            size_core = ef_segy2(coords_c).ravel()
+
+            sp_data_core = np.empty((XX_c.size, 4), dtype=np.float64)
+            sp_data_core[:, 0] = XX_c.ravel()
+            sp_data_core[:, 1] = YY_c.ravel()
+            sp_data_core[:, 2] = ZZ_c.ravel()
+            sp_data_core[:, 3] = size_core
+
+            del XX_c, YY_c, ZZ_c, coords_c, size_core
```
Olender:
```suggestion
```

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+
+            XX_out = XX_p[out_mask]
+            YY_out = YY_p[out_mask]
+            ZZ_out = ZZ_p[out_mask]
+
+            # Padding sizing
+            if extend_segy:
+
+                coords_p = np.column_stack((ZZ_out, XX_out, YY_out))
+                size_pad = ef_segy2(coords_p).ravel()
```
Olender, line 616: WHat if we are reading a binary?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+            else:
+
+                XX_proj = np.clip(XX_out, bbox[2], bbox[3])
+                YY_proj = np.clip(YY_out, bbox[4], bbox[5])
+                ZZ_proj = np.clip(ZZ_out, bbox[0], bbox[1])
+
+                coords_p = np.column_stack((ZZ_proj, XX_proj, YY_proj))
+                base_size_pad = ef_segy2(coords_p).ravel()
+
+                tx = (
```
Olender, line 626: What does t mean?

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+            sp_data_pad = np.empty((XX_out.size, 4), dtype=np.float64)
+            sp_data_pad[:, 0] = XX_out
+            sp_data_pad[:, 1] = YY_out
+            sp_data_pad[:, 2] = ZZ_out
+            sp_data_pad[:, 3] = size_pad
+
+            del XX_p, YY_p, ZZ_p, XX_out, YY_out, ZZ_out
+            if not extend_segy:
+                del XX_proj, YY_proj, ZZ_proj, base_size_pad, tx, ty, tz, t
+            del coords_p, size_pad
```
Olender:
```suggestion
```

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+                t = np.clip(t, 0.0, 1.0)
+                size_pad = base_size_pad + t * (h_padding - base_size_pad)
+
+            sp_data_pad = np.empty((XX_out.size, 4), dtype=np.float64)
+            sp_data_pad[:, 0] = XX_out
+            sp_data_pad[:, 1] = YY_out
+            sp_data_pad[:, 2] = ZZ_out
+            sp_data_pad[:, 3] = size_pad
+
+            del XX_p, YY_p, ZZ_p, XX_out, YY_out, ZZ_out
```
Olender:
```suggestion
```

spyro/meshing/meshing_gmsh3D.py
```diff
@@ -0,0 +1,876 @@
...
+            gmsh.model.mesh.field.setAsBackgroundMesh(1)
+            # Parallel HXT in 3D
+            gmsh.option.setNumber("General.NumThreads", np.int64(num_threads/2))
+            gmsh.option.setNumber("Mesh.MaxNumThreads3D", np.int64(num_threads/2))
+            # Serialize the boundary mesh generation for deterministic results
+            gmsh.option.setNumber("Mesh.MaxNumThreads1D", 1)
+            gmsh.option.setNumber("Mesh.MaxNumThreads2D", 1)
+
+
+def apply_structured_winslow_smoothing3D(
```
Olender, line 678: Half of these paramters are unused

Olender:
I really like the improvements, the code is a lot more readable

spyro/meshing/meshing_utils3D.py
```diff
@@ -0,0 +1,678 @@
...
+        minimum = float(np.min(sizes))
+        raise ValueError(
+            "The structured Winslow sizing function must be positive; "
+            f"minimum value is {minimum}."
+        )
+
+    return sizes
+
+
+def _read_velocity_binary3D(
```
Olender: could you move this method to the io folder?

Olender: Velocity parameter io operations should be there, so other devs can find and use them

rasjr1305: Yes I will move there nice catch

spyro/meshing/meshing_gmsh3D.py
rasjr1305: Will search for a solution that would not break the running grid performance.

spyro/meshing/meshing_functions.py
rasjr1305: fixed

spyro/meshing/meshing_gmsh3D.py
rasjr1305: Will let user choose the number of threads, if user does not specify it will be set to 1

spyro/meshing/meshing_gmsh3D.py
rasjr1305: fixed

spyro/meshing/meshing_gmsh3D.py
rasjr1305: The reader is taking care of that, even if the user inputs other axis order, this error is in case any of those comes wrong from the reader

spyro/meshing/meshing_gmsh3D.py
rasjr1305: fixed

spyro/meshing/meshing_gmsh3D.py
rasjr1305: in this part the binary is already readed and interpolated to ef segy, in the future we might be able to input analytical functions

spyro/meshing/meshing_gmsh3D.py
rasjr1305: These are linear aproximations if the padding size is selected constant, so the padding maximum value is at the far side and slowly shrinks to the mesh border size

spyro/meshing/meshing_gmsh3D.py
rasjr1305: fixed

spyro/meshing/meshing_gmsh3D.py
rasjr1305: fixed

spyro/meshing/meshing_gmsh3D.py
rasjr1305: we need to set gmsh.option.setNumber("Mesh.MaxNumThreads2D", 1 to make the parallel mesh deterministic

spyro/meshing/meshing_utils3D.py
rasjr1305: Moved to basic io

PR #363, opened 2026-08-11
DO NOT SQUASH: Refactor error metrics (6) 

ruansava:
This pull request refactors error calculation throughout the codebase to consistently use the `MeasureError` class from `spyro.tools.error_measure`, replacing custom error calculation functions. It also extends and improves the `MeasureError` class, adding new methods and documentation. The integration and analytical tests are updated to use the new error measures, increasing robustness and clarity of test assertions.

**Refactoring and Consistency Improvements:**
* Replaced all custom error calculation functions in tests and core modules with the `MeasureError` class methods (`peak_error`, `integral_error`, and the new `normalized_root_mean_square_error`). This ensures consistency in error computation across the codebase.

**Enhancements to `MeasureError`:**
* Added the `normalized_root_mean_square_error` method for more comprehensive error analysis, and improved documentation for all error-related methods.

**Test Improvements:**
* Updated integration and analytical tests to assert that all error measures (`NRMS`, `integral`, and `peak`) are below strict thresholds, and to print more informative messages using `parallel_print`.


spyro/tools/cells_per_wavelength_calculator.py
```diff
@@ -491,30 +489,17 @@ def error_calc(receivers, analytical, dt):
...
-
-    squared_error = total_numerator / total_denumenator
-
-    error = np.sqrt(squared_error)
+    for i in range(number_of_receivers):
+        ana[:, i] = np.interp(time_vector_rec, time_vector_ana, analytical[:, i])
+
+    error = spyro.tools.error_measure.MeasureError().error_measures(
+        receivers, ana, dt, number_of_receivers, save_file=False)[3]
+
```
Olender, line 505: The old error would interpolate if the time series was different between the reference and numerical solutions. Does the new error measure does this?

ruansava: I only pass the side after interpolation. That is, it only works with a time series like an array

test_integration/test_forward_acoustic.py
```diff
@@ -91,20 +85,23 @@ def test_forward_3_shots():
...
-            flush=True,
-        )
-    error_all = COMM_WORLD.allreduce(error, op=MPI.SUM)
-    error_all /= 3
-
-    test = np.abs(error_all) < 0.01
-
-    assert test
+    # Computing errors
+    errPk = MeasureError().peak_error(arr0[:430], analytical_p[:430])[0]
```
ndf-ti: suggestion: mark MeasureError methods with `@staticmethod`, this way you dont need to instantitate a new `MeasureError` each time

ruansava: peak_error is already a static method. However, for the other measures, I need to check the inputs through a class method. I think the solution is to define a single instance of this beforehand

spyro/tools/cells_per_wavelength_calculator.py
Olender: I apologise. I didn't really understand your answer. What do you mean by side? Also, time_io has an interpolate_time_series method, that I manly use for interpolating recorded 'real' shot records

spyro/tools/error_measure.py
```diff
@@ -170,6 +179,41 @@ def get_reference_signal(self):
 
         return receivers_reference, receivers_ref_fft
 
+    @staticmethod
+    def check_signal_lengths(signal_model, signal_reference):
```
Olender: I think the name is misleading. This method does not check, it adds a buffer to the transient signal

spyro/tools/error_measure.py
```diff
@@ -249,13 +292,8 @@ def integral_error(signal_model, signal_reference, dt):
...
 
         # Completing with zeros if arrays lengths are different
-        model_len = len(signal_model)
-        reference_len = len(signal_reference)
-        delta_len = abs(model_len - reference_len)
-        if reference_len < model_len:
-            signal_reference = concatenate([signal_reference, zeros(delta_len)])
-        elif reference_len > model_len:
-            signal_model = concatenate([signal_model, zeros(delta_len)])
+        signal_model, signal_reference = self.check_signal_lengths(signal_model,
```
Olender: I don't think we should do this since it can mask stuff without the user knowing. For example:

Lets say a pass a reference and a numerical transient time series with different dts, without noticing. Instead of getting an error telling me what I did wrong a 0 pad will be added to the shorter signal and an error result will be shown that isn't necessarily correct.

spyro/tools/error_measure.py
```diff
@@ -388,11 +430,48 @@ def error_measures(self, forward_solution_receivers, receivers_reference, dt,
...
+        """
+
+        # Check the input parameters
+        validate_data_structure("signal_model", signal_model, "array",
+                                expected_type_element="float")
+        validate_data_structure("signal_reference", signal_reference, "array",
+                                expected_type_element="float")
+
+        # Completing with zeros if arrays lengths are different
+        signal_model, signal_reference = self.check_signal_lengths(signal_model,
```
Olender: Same comment as before. I think this method needs a name change that makes it clear it's not checking but adding a buffer/pad to the data.

In my option if both data inputs have different lengths a simple error should be shown.The user gets to decide if he wants to interpolate in time (cases where dts are different but delays are the same) or if the user wants to add a zero buffer in the beginning (what would be the use case for this?).

spyro/tools/cells_per_wavelength_calculator.py
ruansava: In fact, this word was "array"

spyro/tools/error_measure.py
ruansava: The method checks and modifies if the condition is not satisfied. What name do you suggest?

test_integration/test_forward_acoustic.py
```diff
@@ -91,20 +85,24 @@ def test_forward_3_shots():
...
+    pprint(f"NRMS Error for shot {wave.current_sources} is {eNRMS:.4e} and test "
+           f"has passed equals {abs(eNRMS) < 0.01}", comm=comm)
+    pprint(f"Integral Error for shot {wave.current_sources} is {errIt:.4e} and test "
+           f"has passed equals {abs(errIt) < 0.01}", comm=comm)
+    pprint(f"Peak Error for shot {wave.current_sources} is {errPk:.4e} and test "
+           f"has passed equals {abs(errPk) < 0.01}", comm=comm)
+
+    error_rm = COMM_WORLD.allreduce(eNRMS, op=MPI.SUM) / 3.
+    error_it = COMM_WORLD.allreduce(errIt, op=MPI.SUM) / 3.
+    error_pk = COMM_WORLD.allreduce(errPk, op=MPI.SUM) / 3.
```
Olender, line 102: maybe we can have the error measure class deal with comms in another PR

tests/on_one_core/test_pml_2d.py
```diff
@@ -1,22 +1,18 @@
+from pytest import mark
```
Olender: I ran some tests and didn't see any improvement with this change

Olender: Do we have benchmark results showing that this makes a measurable difference to the test suite? I'd prefer to keep pytest.mark unless the import change has a meaningful impact, since the performance difference here should be negligible

spyro/tools/error_measure.py
ruansava: I cannot suppose the shape of the signal, so that at least I should ensure that the signals are the same size. I should not make a comparison of signals with different sample frequencies. This method can show a message about the signal size in relation to the reference. Thus, at least user knows that is making a comparison that migth be wrong

spyro/tools/error_measure.py
ruansava: Ok. The same comment as above

tests/on_one_core/test_pml_2d.py
Olender: This case is different than `from firedrake import *`, which is a wildcard and will add some overhead. `from pytest import mark` still imports pytest the same way `import pytest` would and I would prefer to keep the consistency throughout the codebase

test_integration/test_forward_acoustic.py
ruansava: Ok!

tests/on_one_core/test_pml_2d.py
ruansava: I see the changes in general. That is, the saving per test is negligible but we have at this moment more than 500 tests. Suppose that you save 1s if you have to run in series, you may save 500s (around 8 min).

spyro/tools/error_measure.py
```diff
@@ -170,6 +179,41 @@ def get_reference_signal(self):
 
         return receivers_reference, receivers_ref_fft
 
+    @staticmethod
+    def check_signal_lengths(signal_model, signal_reference):
+        """Check if the lengths of the model and reference signals are equal.
```
Olender: Same as before, the docstrisng is also misleading. This method doesn't check whether the signal lengths are equal. it zero-pads the shorter signal so that both have the same length. I think the name should reflect that behavior (e.g. `equalize_signal_lengths` or `pad_signal_lengths`)

Olender: and the docstring should reflect that

ruansava: Nice! This name is ok. I will change it

spyro/tools/error_measure.py
```diff
@@ -170,6 +179,41 @@ def get_reference_signal(self):
...
+        ----------
+        signal_model : `array`
+            Transient response at the receiver for the model.
+        signal_reference : `array`
+            Transient response at the receiver for the reference model.
+
+        Returns
+        -------
+        signal_model : `array`
+            Modified transient response at the receiver for the model.
```
Olender:
```suggestion
            Transient response at the receiver for the model, modified with a zero pad if shorter than the reference.
```

spyro/tools/error_measure.py
```diff
@@ -170,6 +179,41 @@ def get_reference_signal(self):
...
+        signal_reference : `array`
+            Modified transient response at the receiver for the reference model.
+        """
+
+        # Getting the maximum length
+        max_len = max(len(signal_model), len(signal_reference))
```
Olender:
```suggestion
        max_length = max(len(signal_model), len(signal_reference))
```

spyro/tools/error_measure.py
Olender: Something like:
```
if len(model_signal) != len(reference_signal):
    raise ValueError(...)
```
Or using the validate methods. Silently padding a signal will hide a time-axis/solver/output issue

Olender: > I cannot suppose the shape of the signal, so that at least I should ensure that the signals are the same size. I should not make a comparison of signals with different sample frequencies. This method can show a message about the signal size in relation to the reference. Thus, at least user knows that is making a comparison that migth be wrong

I don't think a message is appropriate, since you are modifying the data. It should just error out so the users decides if he wants to pad or interpolate the series in a different time-axis. Passing different lengths is a user error and just padding it will either mask or add stuff silently

Olender (changes requested):

spyro/tools/error_measure.py
ruansava: I remember that I used this because I had to compare signals of the infinite model with less steps and model with HABCs with full transient response. Thus, if HABCs are ok, I run the infinite model completely. For this reason, I complete with zeros the array. That is, I can compare two signals with different sizes to make decisions and avoid running an expensive model from a computational point of view

tests/on_one_core/test_pml_2d.py
Olender: I really didn't observe this. pytest and python in the same instance will not do the same thing the second time an import samesomething is used

spyro/tools/cells_per_wavelength_calculator.py
Olender: Ahhh perfect! Then I agree, the array should only be passed after interpolation!

spyro/tools/error_measure.py
Olender: I want to be clear, that I like the pad adding method. I just do not want it to be invisible. It makes sense for HABCs in the example you just mentioned. However, I've seen (done by others and by me) tests where error convergence was not what I expected because the number of timesteps was wrong (like the final_time/dt instead of final_time/dt +1 mistake) and this shows up the moment you are comparing it since the lengths wouldn't match with a correct reference signal. Sometimes we are comparisng results done in different meshes, where dts are different, if we just pad the user will think there's an error different than the real error. I've also seen error comparisons with different softwares (like devito) that have different starting delays for the ricker wavelet. All of these cases are different in how you would treat a comparison. For the first you will notice the bug the second the error shows up, the second you will have to interpoalte the time series, the third you would add a pad on the front of the time series

ruansava: I got. We can force the error or not. That is, we can add a flag always true. Some as `error_if_different_length=True`

Olender: Having a padding method is completely reasonable and good to have it related to the error measure. However, since this measure is now going to be re-utilized throughout the codebase it should be more general. For comparing signals, depending on the use-case the users needs to know if he will pad in the beginning of the series, in the end, interpolate into a different time-axis, or do a mix of all of those

ruansava: I think to open an issue to include the interpolation since I have to identify the tests were it is applied. Now, I wiil include the padding as an option

tests/on_one_core/test_analytical_solution.py
```diff
@@ -1,17 +1,11 @@
-import numpy as np
-import pytest
+from pytest import mark
```
Olender:
```suggestion
import pytest
```

Let's keep it consistent to the other tests. There's no reason for this change

tests/on_one_core/test_analytical_solution.py
```diff
@@ -1,17 +1,11 @@
...
 
-def error_calc(p_numerical, p_analytical, nt):
-    norm = np.linalg.norm(p_numerical, 2) / np.sqrt(nt)
-    error_time = np.linalg.norm(p_analytical - p_numerical, 2) / np.sqrt(nt)
-    div_error_time = error_time / norm
-    return div_error_time
-
-
-@pytest.mark.parametrize("use_vertex_only_mesh", [False, True])
+@mark.parametrize("use_vertex_only_mesh", [False, True])
```
Olender:
```suggestion
@pytest.mark.parametrize("use_vertex_only_mesh", [False, True])
```

tests/on_one_core/test_pml_2d.py
Olender:
```suggestion
import pytest
```

tests/on_one_core/test_pml_2d.py
```diff
@@ -92,34 +88,42 @@ def run_forward():
...
     p_r = wave.forward_solution_receivers
 
-    return p_r, nt
+    return p_r, wave.dt
 
 
-@pytest.mark.slow
-@pytest.mark.skip(reason="Ruben is implementing a right PML formulation")
+@mark.slow
+@mark.skip(reason="Ruben is implementing a right PML formulation")
```
Olender:
```suggestion
@pytest.mark.slow
@pytest.mark.skip(reason="Ruben is implementing a right PML formulation")
```

tests/on_one_core/test_time_convergence.py
```diff
@@ -1,14 +1,7 @@
...
-import pytest
-
-
-def error_calc(p_numerical, p_analytical, nt):
-    norm = np.linalg.norm(p_numerical, 2) / np.sqrt(nt)
-    error_time = np.linalg.norm(p_analytical - p_numerical, 2) / np.sqrt(nt)
-    div_error_time = error_time / norm
-    return div_error_time
+from numpy import isclose, load, log
+from pytest import mark
```
Olender:
```suggestion
import numpy as np
import pytest
```

tests/on_one_core/test_time_convergence.py
```diff
@@ -94,8 +87,8 @@ def run_forward(dt, with_pml=False):
     return rec_out
 
 
-@pytest.mark.slow
-@pytest.mark.parametrize("with_pml", [False, True])
+@mark.slow
+@mark.parametrize("with_pml", [False, True])
```
Olender:
```suggestion
@pytest.mark.slow
@pytest.mark.parametrize("with_pml", [False, True])
```

tests/on_one_core/test_time_convergence.py
```diff
@@ -112,20 +105,21 @@ def test_second_order_time_convergence(with_pml):
...
     errors = []
+    measure_error = MeasureError()
 
     for i in range(len(dts)):
         dt = dts[i]
         rec_out = run_forward(dt, with_pml=with_pml)
-        rec_anal = np.load(analytical_files[i])
-        time = np.linspace(0.0, 1.0, int(1.0 / dts[i]) + 1)
-        nt = len(time)
+        rec_anal = load(analytical_files[i])
```
Olender:
```suggestion
        rec_anal = np.load(analytical_files[i])
```

tests/on_one_core/test_time_convergence.py
```diff
@@ -112,20 +105,21 @@ def test_second_order_time_convergence(with_pml):
...
-        errors.append(error_calc(rec_out.flatten(), rec_anal, nt))
+        errors.append(measure_error.normalized_root_mean_square_error(rec_out.flatten(),
+                                                                      rec_anal))
 
     theory = [t**2 for t in dts]
     theory = [errors[0] * th / theory[0] for th in theory]
 
-    assert math.isclose(np.log(theory[-1]), np.log(errors[-1]), rel_tol=3e-2)
+    assert isclose(log(theory[-1]), log(errors[-1]), rtol=3e-2), \
+        "Second order time convergence not achieved."
```
Olender:
```suggestion
    assert np.isclose(np.log(theory[-1]), np.log(errors[-1]), rtol=3e-2),
```

tests/parallel/test_forward.py
```diff
@@ -1,19 +1,13 @@
-import firedrake as fire
-import pytest
+from pytest import mark
```
Olender:
```suggestion
import pytest
```

tests/parallel/test_forward.py
```diff
@@ -1,19 +1,13 @@
...
 
-def error_calc(p_numerical, p_analytical, nt):
-    norm = np.linalg.norm(p_numerical, 2) / np.sqrt(nt)
-    error_time = np.linalg.norm(p_analytical - p_numerical, 2) / np.sqrt(nt)
-    div_error_time = error_time / norm
-    return div_error_time
-
-
-@pytest.mark.parallel(6)
+@mark.parallel(6)
```
Olender:
```suggestion
@pytest.mark.parallel(6)
```

tests/three_dimension/test_hexahedral_convergence.py
```diff
@@ -1,12 +1,8 @@
-import numpy as np
+from firedrake import COMM_WORLD as comm
+from numpy import pi, save, sqrt
```
Olender:
```suggestion
import numpy as np
```

tests/three_dimension/test_hexahedral_convergence.py
```diff
@@ -88,20 +84,19 @@ def run_forward_hexahedral(dt, final_time, offset):
...
     wave.set_initial_velocity_model(constant=1.5)
     wave.forward_solve()
-    # time_vector = np.linspace(0.0, final_time, int(final_time/dt)+1)
 
     rec_out = wave.forward_solution_receivers
     output = rec_out.flatten()
     my_ensemble = wave.comm
     if my_ensemble.comm.rank == 0 and my_ensemble.ensemble_comm.rank == 0:
-        np.save("dofs_3D_quads_rec_out"+str(dt)+".npy", output)
+        save("dofs_3D_quads_rec_out"+str(dt)+".npy", output)
```
Olender:
```suggestion
        np.save("dofs_3D_quads_rec_out"+str(dt)+".npy", output)
```

tests/three_dimension/test_hexahedral_convergence.py
```diff
@@ -88,20 +84,19 @@ def run_forward_hexahedral(dt, final_time, offset):
...
 
     return output
 
 
 def analytical_solution(dt, final_time, offset):
-    amplitude = 1/(4*np.pi*offset)
-    delay = offset/1.5 + 1.5 * np.sqrt(6.0) / (np.pi * 5.0)
+    amplitude = 1 / (4 * pi * offset)
+    delay = offset / 1.5 + 1.5 * sqrt(6.0) / (pi * 5.0)
```
Olender:
```suggestion
    amplitude = 1 / (4 * np.pi * offset)
    delay = offset / 1.5 + 1.5 * np.sqrt(6.0) / (np.pi * 5.0)
```

tests/three_dimension/test_hexahedral_convergence.py
```diff
@@ -119,13 +114,18 @@ def test_3d_hexa_one_source_propagation():
...
-    small_error = error_time < 0.02
+    # Computing errors
+    measure_error = MeasureError()
+    eNRMS = measure_error.normalized_root_mean_square_error(p_numeric, p_analytic)
+    errIt = measure_error.integral_error(p_numeric, p_analytic, dt)
+    errPk = measure_error.peak_error(p_numeric, p_analytic)[0]
 
-    print(f"Error is smaller than necessary: {small_error}", flush=True)
+    assert abs(eNRMS) < 0.02 and abs(errIt) < 0.02 and abs(errPk) < 0.02, \
+        "Error is too high for forward test with hexahedral mesh."
```
Olender:
```suggestion
    assert np.abs(eNRMS) < 0.02 and np.abs(errIt) < 0.02 and np.abs(errPk) < 0.02, \
        "Error is too high for forward test with hexahedral mesh."
```

spyro/tools/__init__.py
```diff
@@ -6,27 +6,3 @@
     "Meshing_parameter_calculator",
     "smooth_velocity_field_file",
 ]
-
-# from .grid_point_calculator import wave_solver, generate_mesh, error_calc
```
Olender, line 10: Good idea removing these commented out lines

test_integration/test_forward_acoustic.py
ruansava: OK!

tests/parallel/test_forward_supershot.py
```diff
@@ -92,19 +83,35 @@ def test_forward_supershot():
...
+    eNRMS = error0_nrms + error199_nrms
+    error_rm = COMM_WORLD.allreduce(eNRMS, op=MPI.SUM) / 2
+    errIt = error0_it + error199_it
+    error_it = COMM_WORLD.allreduce(errIt, op=MPI.SUM) / 2
+    errPk = error0_pk + error199_pk
+    error_pk = COMM_WORLD.allreduce(errPk, op=MPI.SUM) / 2
 
-    if comm.comm.rank == 0:
-        print(f"Combined error for shots {wave.current_sources} is {error_all} and test has passed equals {np.abs(error_all) < 0.01}", flush=True)
+    comm.comm.barrier()
```
Olender, line 104: Does this break without this barrier? I'm not against having it, I just want to know (sometimes we just need a barrier, but I imagined it wouldn't be necessary just after an allreduce operation

Olender (approved):

tests/parallel/test_forward_supershot.py
ruansava: I don't modify the code in this level. I only modified/added the measure errors according to the current structure

PR #368, opened 2026-08-13
docs: update README mesh setup example

### LINKED ISSUE #291: Main GitHub page example error
The example in the main GitHub page is returning an error due the out dated setting of mesh size

Image

I believe it now should work by setting the edge lenght

Wave_obj.set_mesh(input_mesh_parameters={"edge_length": 0.01})

or 

```
# Define the domain size without the PML. Here we'll assume a 0.75 x 1.50 km
dictionary["mesh"] = {
    # depth in km - always positive
    "length_z": 0.75,
    # width in km - always positive
    "length_x": 1.50,
    # thickness in km - always positive
    "length_y": 0.0,
    # If we are loading and external .msh mesh file
    "mesh_file": None,
    # options: None (default), firedrake_mesh, user_mesh, or SeismicMesh
    # use this opion if your are not loading an external file
    # 'firedrake_mesh' will create an automatic mesh using firedrake's built-in meshing tools
    # 'user_mesh' gives the option to load other user generated meshes from unsuported formats
    # 'SeismicMesh' automatically creates a waveform adapted unstructured mesh to reduce total
    # DoFs using the SeismicMesh tool.
    "mesh_type": "firedrake_mesh",
    "edge_length": 0.02, # <--------
}
```

floze-the-genius:
## Summary

- update the README's automatic-mesh example to use the current `set_mesh` API
- pass `edge_length` through `input_mesh_parameters` instead of the removed `dx` keyword

## Validation

- `git diff --check`
- verified the documented keyword and nested `edge_length` parameter against the current `Wave.set_mesh` and `MeshingParameters` implementations
- Firedrake demo not run locally because the Firedrake runtime is unavailable on macOS

Closes #291

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,7 @@ dictionary["visualization"] = {
 wave = spyro.AcousticWave(dictionary=dictionary)
 
 # Defines the element size in the automatically generated firedrake mesh.
-wave.set_mesh(dx=0.01)
+wave.set_mesh(input_mesh_parameters={"edge_length": 0.01})
 
 
 # Manually create a simple two layer seismic velocity model.


Olender:
Hello @floze-the-genius and welcome to the spyro repository. We still haven't added our AI contribution policy, so before I approve I wanted to ask: did you use an AI tool and if so, which? We will not accept AI-written PRs that fix issues labelled 'good first issue'. These issues are good entry points for newcomers to spyro and open source software so having them be closed by an AI tool is not really beneficial.

Olender:
And another thing. In your comment you said Firedrake was not available in MacOS. But I want to point out that for ARM-based MACs (which are includes every mac released in the past 6 years) it is available (a significant number of Firedrake devs actually have macbooks). Not that you need to install firedrake for this issue, but if you want to have a look at it, the installation instructions are available at: https://www.firedrakeproject.org/install.html

PR #370, opened 2026-08-14
Add proper time axis and acquisition model classes

ndf-ti:
Created the following objects:

- `TimeAxis': This class will be the source of truth for time related variables, such as dt, final and initial time. It will also contain the current time of the simulation as well as the index of the iteration
- `Position`: This is a utility class for when we need to store positions. I have seem some pieces of code where we check the dimension and if the dimension is equal 3, y is considered. With this new class i want to avoid these situations. The `get_positions` will return a tuple containing (x, y) or (x, y, z) depending on the obejct initialization. So if you need to pass coordinates arguments to another function, just pass this tuple instead without doing dimensions check. Example:
```Python
p = Position(0, 0)

mesh.locate_cell(p.get_position()) # no need to check for dimensions
```
- `Receiver`: This class will represent a receiver. It has methods to sample the wave function as well as calculating the misfit
- `Source`: This class represents a source. The update method can change the function values so they can be injected in the wave field while doing FWI. There are 2 subclasses, one for a ricker source and another for custom data.
- `Acquisition`: This class will represent the acquisition geometry. It has a method to provide the source function, sample data at the receiver and to calculate the misfit.

Also added unit tests for each class.

With this structure, i am thinking in creating another class for the FWI loop like:

```python

time = TimeAxis(...)
acquisition = Acquisition(...)

forward = ForwardFormulation(acquisition,....)

while time.can_update():
     u = forward.solve()
     acquisition.update_sources(time)
     acquisition.store_sample_at_receivers(u, time)
     time.update()

misfit = acquisition.get_misfit()
```

Let me know if you guys think this structure makes sense!


spyro/model/acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+import firedrake as fire
+from spyro.model.position import Position
+from spyro.model.time_axis import TimeAxis
+
+
+@dataclass
+class Receiver:
+    p: Position
+    real: np.array
+    mesh: InitVar[fire.mesh.MeshGeometry]
```
Olender: I think the acquisition geometry input should come before the mesh generation(since we can use the same acquisition geometry for different meshes). Therefore, it shouldn't have the fire.mesh object

spyro/model/acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+    synthetic: np.array = field(init=False)
+    point_evaluator: fire.PointEvaluator = field(init=False)
+
+    # Adicionar validação de dominio
+    def __post_init__(self, mesh: fire.mesh.MeshGeometry):
+        self.synthetic = np.zeros_like(self.real)
+        self.point_evaluator = fire.PointEvaluator(mesh, [self.p.get_position()])
```
Olender: This part needs to be left up to the solver, since we need to make sure it's annotatable if thats the case

spyro/model/acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+    def get_misfit(self):
+        residual = self.synthetic - self.real
+
+        return 0.5 * np.sum(residual**2)
+
+
+@dataclass
+class Source(ABC):
+    p: Position
+    space: InitVar[fire.FunctionSpace]
```
Olender: My previous observation regarding the mesh obj is also valid for the FUnctionSpace (which is created after the mesh)

spyro/model/acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def __post_init__(self, space: fire.FunctionSpace):
+        mesh = space.mesh()
+
+        source_mesh = fire.VertexOnlyMesh(
+            mesh,
+            [self.p.get_position()],
+        )
+
+        source_space = fire.FunctionSpace(
```
Olender: The source_space should be defined somewhere else (since it needs a mesh and will be different for different wave propagators). I think this class should handle only inputs instead of solver related things. We need to keep the boundary between inputs and discretization

spyro/model/acquisition.py
```diff
@@ -0,0 +1,137 @@
...
+@dataclass
+class DataSource(Source):
+    data: np.array
+
+    def amplitude(self, time: TimeAxis) -> float:
+        return self.data[time.index]
+
+
+@dataclass
+class RickerSource(Source):
```
Olender: This is nice!

spyro/model/time_axis.py
```diff
@@ -0,0 +1,29 @@
+from dataclasses import field, dataclass
+import warnings
+
+
+@dataclass
+class TimeAxis:
```
Olender: Perfect

Olender:
I have a major concern about this PR: you are coupling acquisition geometry with mesh/discretization. I think dataclasses are very good for our input layer (and will make excelent substitutes to the modelparameters_io.py stuff). But there should be a hard boundary between input data classes and Firedrake runtime objects. Pure io spyro stuff can be validated without expensive Firedrake objects and functions.

A multiscale FWI algortihm runs something like this in spyro:

validate inputs and configuration
      ↓
POINT A
      ↓
generate mesh
      ↓
create FunctionSpace 
      ↓
    POINT B
          ↓
    run forward, run gradient
          ↓
    update model
          ↓
    RETURN TO POINT B until some condition is met,
      ↓
RETURN TO POINT A, until some condition is met
      ↓
generate outputs.

Without multiscale FWI (a simple FWI), we can just take out the return to point A step.

Olender:
Both the source and receiver input classes shouldn't have access to FunctionSpace, Cofunctions. In our workflow the acquisition geometry is independent of the numerical discretization: we validate the inputs, generate a mesh and numerous function spaces, and then solve the problem. In particular, the mesh (and the functionspaces dependent on it) can be regenerated during FWI/remeshing while the sources and receivers remain unchanged. Because of that, I'd prefer Source/Receiver/Acquisition to represent the acquisition itself, without owning mesh or function-space-dependent Firedrake objects.

Olender:
I really like the move to data classes, but we need to keep our current boundary between inputs and numerical discretization intact

Olender (changes requested):

spyro/model/acquisition.py
ndf-ti: Fair. This was meant to just be a tool to validate if the receiver is inside the mesh. I think i can move the validation to another point.

spyro/model/acquisition.py
ndf-ti: what do you mean with annotatable?

spyro/model/acquisition.py
Olender: It can be validated using mesh dimensions, without creating a mesh object

spyro/model/acquisition.py
Olender: Any solver-related operation has to be annotatable. To create automatic adjoints every operator has to be recorded on a tape and the backward propagator will basically transpose these operators and apply them backwards (this process is more complex than this and @Ig-dolci understands it a lot better than me if you want a better explanation). However, we do not necessarily use annotation (for the implemented adjoint I have it turned off). So these solver-related operations have to be created after we define if annotation is on or off.

Ig-dolci:
Me parece que estamos complicando sources e receivers nesse PR. Preciso entender melhor o objetivo desse PR numa conversa com o Felipe.

Olender:
> Me parece que estamos complicando sources e receivers nesse PR. Preciso entender melhor o objetivo desse PR numa conversa com o Felipe.

Então, acredito que essa estrutura de classes de dados, bem empregadas e depois de revisão com os outros devs (pois engloba inputs que todos usam), podem simplficiar o `model_parameters.py`  que atualmente está uma bagunça (culpa minha), possivelmente até substituindo ele (mantendo um wrapper pra usuários que ainda preferem o uso de dictionários de entrada).

Mesmo que não vierem a `substituir o model_parameters.py` ainda vamos precisar dela (mesmo que acessadas apenas por wrappers) para integrar melhor com a estrutura de software da Total (inclusive isso foi pedido por eles).

Temos que conversar mais com o Filipe, pois isso seria para lidar com os inputs (validação, etc.) e não para lidar com a parte de discretização, solvers, vertexonlymesh, etc.

PR #371, opened 2026-08-14
Viscoelastic isotropic wave propagation

PNAraujo:


spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+
+    Examples
+    --------
+    ``(ElasticMaterialParameter.DENSITY, ElasticMaterialParameter.MU)``
+    becomes ``"{density, mu}"``.
+    """
+    return "{" + ", ".join(parameter.value for parameter in parameters) + "}"
+
+
+class AnisotropicTTIWave(ElasticWave):
```
Olender:
```suggestion
class AnisotropicTTIWave(IsotropicWave):
```

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
+import numpy as np
+
+from firedrake import (assemble, Constant, curl, DirichletBC, div, Function,
+                       FunctionSpace, project)
+
+from .elastic_wave import ElasticWave
```
Olender:
```suggestion
from .elastic_wave import IsotropicWave
```

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+
+class AnisotropicTTIWave(ElasticWave):
+    '''Anisotropic elastic wave propagator'''
+
+    def __init__(self, dictionary, comm=None):
+        super().__init__(dictionary, anisotropy = WaveType.ANISOTROPIC_TTI_ELASTIC, comm=comm)
+        self.rho = None   # Density
+        self.lmbda = None  # First Lame parameter
+        self.mu = None    # Second Lame parameter
+        self.c_s = None   # Secondary wave velocity
```
Olender:
```suggestion
```
If we inherit from IsotropicWave we can remove duplicated code

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+        self.delta = None
+        self.epsilon = None
+        self.gamma = None
+        self.theta = None
+        self.phi = None
+        self._control_parameterization = None
+        self._material_parameter_function_space = None
```
Olender:
```suggestion
```
Same as previous comment

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+
+        # Variables for logging the S-wave
+        self.s_wave = None
+        self.C_h = None
+        self.field_logger.add_field("s-wave", "S-wave",
+                                    lambda: self.update_s_wave())
+
+        self.mechanical_energy = None
+        self.field_logger.add_functional("mechanical_energy",
+                                         lambda: assemble(self.mechanical_energy))
```
Olender:
```suggestion
```
Same as previous comment

spyro/solvers/elastic_wave/forms.py
```diff
@@ -53,6 +52,180 @@ def isotropic_elastic_without_pml(wave):
         lin_var, solver_parameters=solver_parameters
     )
 
+def elastic_without_pml(wave):
```
Olender, line 55:
```suggestion
def anisotropic_elastic_without_pml(wave):
```
We should reinstate the previous form and remane this one. What do you think?

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+                    self.mesh, self.method, self.degree, dim=1,
+                )
+                return Function(V).interpolate(value)
+            return value
+
+        def get_value(parameter, *aliases):
+            for key in (parameter.value, *aliases):
+                if key in synthetic_data_dict:
+                    return material_parameter(synthetic_data_dict[key])
+            return None
```
ndf-ti, line 131: why use nested functions?

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+        ``{DENSITY: rho, P_WAVE_VELOCITY: c, S_WAVE_VELOCITY: c_s}``.
+        """
+        parameterization = self._control_parameterization
+        if parameterization is None:
+            if self.rho is None:
+                return None
+            parameterization = ElasticMaterialParameterization.LAME
+
+        parameters = {}
+        for parameter in CONTROL_PARAMETERS_BY_PARAMETERIZATION[parameterization]:
```
ndf-ti, line 212: why not just set the parameters explicitly? like:

```python
parameters: {
    DENSITY: self.rho,
    LAMBDA: self.lmbda,
    ...
}
```

Olender:
Hi @PNAraujo , is this PR dependent on PR #378 ?

PNAraujo:
Hi, @Olender! No, this PR does not depend on PR #378. It solves the anisotropic elastic wave equation in a different way, but I can check whether viscoelasticity can be incorporated directly into this structure.

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+            Examples
+            --------
+            ``density=1.0`` becomes ``Constant(1.0)`` before the mesh exists,
+            and becomes a scalar material ``Function`` after the mesh has been
+            created.
+            """
+            if np.isscalar(value) or isinstance(value, Constant):
+                if self.mesh is None:
+                    return Constant(value) if np.isscalar(value) else value
+                V = create_function_space(
```
Olender, line 121: Is there a previously created scalar function space we could point to?

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+        displacement solution lives in a vector function space. This helper
+        keeps the inversion controls in the scalar space returned by
+        ``get_control_parameter_function_space()`` so density, Lame
+        parameters, and velocity controls can be flattened, rebuilt, written,
+        and reassigned consistently during FWI.
+
+        Accepted values are Firedrake ``Function`` objects, constants, scalar
+        values, or UFL expressions. Functions already in the target space are
+        copied with ``assign``; all other values are interpolated into a new
+        named scalar ``Function``.
```
Olender: Very clear docs, nice!

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+            self.lmbda = self.rho*self.c**2 - 2*self.mu
+            self._control_parameterization = ElasticMaterialParameterization.VELOCITY
+            synthetic_data["p_wave_velocity"] = self.c
+            synthetic_data["s_wave_velocity"] = self.c_s
+            synthetic_data["delta"] = self.delta
+            synthetic_data["epsilon"] = self.epsilon
+            synthetic_data["gamma"] = self.gamma
+            synthetic_data["theta"] = self.theta
+            synthetic_data["phi"] = self.phi
+            synthetic_data["anisotropy"] = self.anisotropy_type
```
Olender, line 365: Are the control parameters set like this with elastic as well? If so, okay, we can open an issue and refactor later

Olender:
Hi, just an overall comment about imports:

I noticed there are a few wildcard imports (`from ... import *`). I generally prefer to avoid them altogether.  Besides unnecessarily polluting the namespace (which can make debugging much more difficult) in this case specially they also hide the actual import structure and dependencies.

For example, in `anisotropic_TTI_wave.py` you have:

```python
from .tensor_computation import *
```

and then in `tensor_computation.py`, you have:

```python
from .anisotropy import *
```

In `anisotropic_TTI_wave.py`, it looks like you only actually only use `C_computation` from `tensor_computation`, so you could instead write:

```python
from .tensor_computation import C_computation
```

However, if you make that change, you'll notice that `WaveType` becomes undefined. That's because `WaveType` is imported into `tensor_computation.py` from `anisotropy.py`, and the wildcard import was effectively making it available transitively in `anisotropic_TTI_wave.py`.

This is a good example of why wildcard imports can make the code difficult to follow. A reader looking at `anisotropic_TTI_wave.py` has no obvious way of knowing where `WaveType` actually comes from, and it becomes harder to trace dependencies when debugging or refactoring. Something very similar also happens inside your `tensor_computation.py` file (where you ger the firedrake as fire import from ´anisotropy.py´)

I'd recommend avoiding wildcard imports here and explicitly importing the names that are actually used. It makes the dependency structure much clearer and avoids relying on transitive imports.

spyro/solvers/elastic_wave/anisotropic_TTI_wave.py
```diff
@@ -0,0 +1,625 @@
...
+                       FunctionSpace, project)
+
+from .elastic_wave import ElasticWave
+from .forms import (elastic_without_pml,
+                    isotropic_elastic_with_pml)
+from .functionals import mechanical_energy_form
+from ...utils.typing import (ElasticMaterialParameter, ElasticMaterialParameterization,
+                             AbsorbingBCsType, override)
+from ...domains.space import create_function_space
+from .tensor_computation import *
```
Olender:
```suggestion
from .anisotropy import WaveType
from .tensor_computation import C_computation
```

spyro/solvers/elastic_wave/tensor_computation.py
```diff
@@ -0,0 +1,346 @@
+from firedrake import (assemble, Constant, curl, DirichletBC, div, Function,
+                       FunctionSpace, project, VectorFunctionSpace, TensorFunctionSpace, sym, grad,
+                       as_matrix, as_vector)
```
Olender:
```suggestion
import firedrake as fire
from firedrake import as_matrix
```

spyro/solvers/elastic_wave/tensor_computation.py
```diff
@@ -0,0 +1,346 @@
...
+
+from ufl import conditional, lt
+import ufl
+from ...utils.typing import  WaveType
+
+from .anisotropy import *
```
Olender:
```suggestion
```

Olender:
We should merge main into here now that PR #382 has merged, so you can use the new physical parameters

spyro/solvers/time_integration_central_difference.py
```diff
@@ -113,27 +113,24 @@ def _propagate_forward_central_difference(wave, source_ids):
...
-
-                # Update memory variables
-                for i in range(len(zeta_list)):
-                    zeta_old = Function(W)
-                    zeta_old.assign(zeta_list[i])
-                    omega = omega_list[i]
-                    zeta_np1 = project(zeta_old + dt * omega * (eps - zeta_old), W)
-                    zeta_list[i].assign(zeta_np1)
+            # Strain rate
+            eps = project(epsilon(wave.vstate), W)
```
Olender, line 127: We should look into this later. Any projection inside the time marching loop can dominate our computational cost

.virtual_documents/notebook_tutorials/viscoelastic_wave_propagation.ipynb
Olender, line 1: Something weird is going on with this notebook

Olender:
I really like having new notebooks, demos, and/or .py.rst files. However, I think you need to rerun the blocks in your notebook. It starts with an import error, but than runs using that module in the end. I think using the `Clear all outputs` - > `Restart` -> `Run all` commands in order should fix this before doing a new git push

PR #372, opened 2026-08-14
Added inital agents file

### LINKED ISSUE #354: Create AGENTS.md
We should create an `AGENTS.md` file with guidelines for coding agents contributing to this repository. The [Firedrake `AGENTS.md`](https://github.com/firedrakeproject/firedrake/blob/main/AGENTS.md) could serve as a useful starting point.

Olender:
Let us define how AI agents should be used within our code.

This is a draft for discussion within the core dev teambased on @Ig-dolci 's issue #354 . @ruansava , @Ig-dolci , @PNAraujo , @rasjr1305  , @j-salesps , @ndf-ti , @raousama3 , and @hassan-imran-ee I would really appreacite your inputs.  I've added an initial file based on the one over at the Firedrake repo. Let us modify it to our specifications


Olender:
How do you guys use AI agents in spyro development? I usually only use them for docs, formatting PR comments, and sometimes some tests (albeit very rarely).

Ig-dolci (approved):

AGENTS.md
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self):
+        self._initialized = False
+
+    def solve(self, pc, b, x):
+        if not self._initialized:
+            self._ksp = PETSc.KSP().create(comm=pc.comm)
+            self._ksp.setOperators(*pc.getOperators())
+            self._initialized = True
+        self._ksp.solve(b, x)
+```
```
ndf-ti: i dont think this is a good pattern. Ideally the setup should be made only one time in the constructor and then the methods should trust that all the properties are initialized. If you are in a position where the object has a dependency that will only be created later in the workflow, then that probably means that the abstraction is not ideal.

Ig-dolci:
I think it is a good starting point. We can add more points in future as needed.

AGENTS.md
Olender: I like the initialization, but I'm going to remove it and we can discuss and add again (or something similar) in the future

PR #373, opened 2026-08-14
DO NOT SQUASH: Introducing AbsorbingBC class (9) 

ruansava:
This pull request introduces an abstract class for all methods of ABCs  (HABCs, PML, and NRBCs).
Thus, it centralizes the operations that are common to all methods.


spyro/abc/abc.py
```diff
@@ -0,0 +1,635 @@
...
+from ..domains.space import create_function_space
+from ..io.basicio import parallel_print as pprint
+from ..tools.abc_labeling_cases import formatting_abc_layer_type
+from ..tools.habc_tools import clipping_coordinates_lay_field, extend_scalar_field_profile
+from ..utils.error_management import (validate_data_structure, validate_numeric,
+                                      validate_parameter)
+from ..utils.typing import LayerShapeType
+
+
+class AbsorbingBC(MeasureError, metaclass=ABCMeta):
```
ndf-ti, line 17: question: why inherit from MeasureError?

spyro/abc/abc.py
```diff
@@ -0,0 +1,635 @@
...
+        -------
+        None
+        """
+
+        # Model dimension
+        self.dimension = validate_parameter("dimension", dimension, [2, 3])
+
+        # Original domain dimensions
+        self.domain_dim = validate_data_structure("domain_dim", domain_dim, "tuple",
+                                                  expected_type_element=("float", "int"),
```
ndf-ti, line 121: dimension can be inferred from domain_dim, right? I think we can set dimension in the constructor, but we dont need to receive it as a parameter.

spyro/abc/abc.py
ruansava: Because it has all the methods to evaluate the efficacy/efficiency of ABCs that I proposed in a paper. Initially, they were methods inside HABC class. Then, we saw that these error measures can be applied to other parts of the code independently. This, I decoupled in a new class and applied in the code where needed

spyro/abc/abc.py
ruansava: Along the code we have always used dimension even though we can infer from the other parameters. However, there are the attribute wave.dimension and generally we are referring to it. That is, we pass it as an input argument

PR #375, opened 2026-08-17
Marmousi with newer gmsh

Olender:
Added newer marmousi based fwi demos


marmousi_forward.py
```diff
@@ -0,0 +1,69 @@
+import spyro
```
ndf-ti, line 1: question: why these files are not in a specific folder?

Olender: You are right import spyro should be below the other imports

PR #376, opened 2026-08-17
fixing docs in sources and receivers

Olender:
Adding numpy style docs to receiver and source related files

I'm not really focues on refactoring these classes or doing any major changes, since they will eventually be completly replaced by vertexonlymesh. However, I wanted to update the docs to follow numpy style doc style


ndf-ti (approved):

krober10nd:
still waiting on vertex only mesh implementation in Firedrake for sources and receivers?! :) that was 5 years ago!

PR #382, opened 2026-08-19
physical parameter in Wave and Control Parameter for FWI/Gradient

Ig-dolci:
# Separate the wave equation's physical parameters from the inversion's control parameters

## Why

`Wave` and `FullWaveformInversion` were sharing one vocabulary for two different
things. The solver exposed `get_control_parameters()`, `set_control_parameters()`
and `get_control_parameter_function_space()` — so a wave equation had to know
what an inversion control was, even though nothing in the forward problem cares.
The two concepts are not the same:

- **Physical parameters** are the material fields the variational form is
  written in terms of. An acoustic medium has one (the velocity model); an
  isotropic elastic medium has five, of which three are independent.
- **Control parameters** are the subset an inversion treats as unknowns. That is
  a property of the optimization, not of the physics.

## What changed

**`Wave` now only knows about physical parameters.**

```python
wave.physical_parameters                    # set of material parameters -> their fields
wave.initialize_physical_parameters()       # build them without a forward solve
AcousticWave._physical_parameter_names      # frozenset(AcousticMaterialParameter)
IsotropicWave._physical_parameter_names     # frozenset(ElasticMaterialParameter)
```

Solvers declare their parameters and register the fields where they already
computed them. No control-parameter API remains on any solver.

**`FullWaveformInversion` owns the controls.** They are held as a single
`PhysicalParameters` mapping each controlled parameter to its value, so *which*
parameters are inverted for is recorded by its members and *their values* by the
fields — there is no second record to fall out of sync. The rule tying the two
layers together holds structurally: the controls are copied out of the solver's
physical parameters, so they cannot fail to be a subset of them.

**New `PhysicalParameters` container** (`spyro/utils/physical_parameters.py`) is
a `collections.abc.Set` over material parameters that also maps each to its
field, so `controls <= wave.physical_parameters` reads directly. Its `update()`
writes *into* the existing `Function` rather than replacing it, which is what
keeps assembled forms — and parameters computed from others, such as the elastic
wave speeds under a Lame parameterization — valid without being rebuilt.

**New `AcousticMaterialParameter` enum** so an acoustic medium names its velocity
without borrowing `ElasticMaterialParameter`. Parameters are keyed by enum
members, not strings.

## Bug fixed

Setting the true and the guess model in sequence — the standard synthetic-FWI
setup, both on the same solver — captured the wrong values:

```python
fwi.set_real_velocity_model(constant=3.0)
fwi.set_guess_velocity_model(constant=2.0)
# guess was captured as 3.0
```

`set_initial_velocity_model` replaced `wave.initial_velocity_model` but left
`wave.c` and the registered physical parameters bound to the previous field, so
the second capture read back the first model. The inversion would have started
from the true model with a near-zero misfit. It is symmetric: in the other
order, the real model was captured as the guess.

Fixed at the root, in `Wave.set_initial_velocity_model`: replacing the model
drops the registered physical parameters, so the next reader rebuilds them.
Covered by `test_velocity_model_setters_capture_each_model_separately`,
parameterized over both orders and confirmed to fail without the fix.

## Public API

**Removed** — `Wave.get_control_parameters`, `Wave.set_control_parameters`,
`Wave.get_control_parameter_function_space`, `FullWaveformInversion.guess_control`
and `.real_control` attributes, `supported_wave_types` and the
non-acoustic guard, `control_result`.

**Added** — `Wave.physical_parameters`, `Wave.initialize_physical_parameters`,
`FullWaveformInversion.control_parameters`, `spyro.PhysicalParameters`,
`spyro.AcousticMaterialParameter`.

**Unchanged** — `set_guess_control(velocity)` still takes and
`control_parameters` still returns a plain `Function` for acoustic inversion.
The dictionary is internal.

`FullWaveformInversion` no longer refuses a non-acoustic solver at construction;
it now fails later, at `gradient_solve`, with *"Elastic adjoint gradients are not
implemented yet"*.

## Testing

`tests/on_one_core/test_fwi_controls.py` rewritten against the new API: 21 tests
covering the physical-parameter set, in-place updates, the control/physical
subset rule, both model-setter orders, and rejection of empty or unmodelled
parameters.

Full non-gradient suite: **483 passed, 17 skipped**. Twenty pre-existing failures
in `test_modal_solvers` (SLEPc not installed), `test_reference_model_abc`,
`test_habc_tools`, `test_eikonal` and `test_gmsh2d*` were verified identical
test-for-test on a worktree at the base commit, so they are not from this change.


spyro/solvers/elastic_wave/isotropic_wave.py
```diff
@@ -193,6 +198,12 @@ def get_value(parameter, *aliases):
                 "The valid options are {Density, Lame first, Lame second} "
                 "or (exclusive) {Density, P-wave velocity, S-wave velocity}",
             )
+        add = self._physical_parameters.add
+        add(ElasticMaterialParameter.DENSITY, self.rho)
+        add(ElasticMaterialParameter.LAMBDA, self.lmbda)
+        add(ElasticMaterialParameter.MU, self.mu)
+        add(ElasticMaterialParameter.P_WAVE_VELOCITY, self.c)
+        add(ElasticMaterialParameter.S_WAVE_VELOCITY, self.c_s)
```
Olender, line 206: Gostei

spyro/solvers/inversion.py
Olender, line 1: Esse arquivo está ficando confuso

spyro/solvers/inversion.py
```diff
@@ -404,39 +430,6 @@ class when ``wave`` is not provided.
...
-            for wave_type in self.supported_wave_types
-        )
-        raise NotImplementedError(
-            "FullWaveformInversion currently supports only acoustic wave "
-            "solvers because adjoint-based gradients are implemented only for "
-            f"{supported_names}. Received {self.wave_type.name}.",
-        )
-
     def _sync_wave_real_shot_record(self):
         """Copy observed data from the FWI driver to the wave solver.
```
Olender, line 407: Is this a copy or a point to?

spyro/solvers/inversion.py
```diff
@@ -454,60 +447,300 @@ def _sync_wave_real_shot_record(self):
...
+        if items is not None:
+            values = dict(items)
+            unknown = set(values) - set(self._control_parameters)
+            if unknown:
+                raise ValueError(
+                    f"{_names(unknown)} are not controlled by this "
+                    f"inversion. Controls: {_names(self._control_parameters)}.",
+                )
+            return values
+        return {self._control_parameters[0]: control}
```
Olender: WHat happens here if control is not None but is empty?

spyro/solvers/elastic_wave/isotropic_wave.py
```diff
@@ -12,7 +12,7 @@
 from ...domains.space import create_function_space
 
 
-CONTROL_PARAMETERS_BY_PARAMETERIZATION = {
+PHYSICAL_PARAMETERIZATION = {
```
Olender, line 15: Is this dictionary still needed now?

Olender:
@PNAraujo , o que tinhámos conversado hoje, talvez esteja sendo resolvido aqui. O que você acha das mudanças nos arquivos `isotropic_wave.py` e `physical_parameters.py` pro seu draft PR?

spyro/solvers/inversion.py
```diff
@@ -31,6 +32,22 @@
 # ROL = None
 
 
+def _names(parameters):
```
ndf-ti: maybe it is better to move this function to another file as it doesn't seem to be part of the inversion

spyro/solvers/inversion.py
```diff
@@ -454,60 +435,223 @@ def _sync_wave_real_shot_record(self):
...
-            return control
-        raise TypeError(
-            "FWI control must be a firedrake Function or Constant. "
-            f"Received {type(value).__name__}.",
+        if control is None:
+            return {}
+        try:
+            items = control.items()
+        except AttributeError:
+            items = None
```
ndf-ti, line 563: i think that instead of assigning items = None here, you can return `{self._controlled_parameters()[0]: control}` directly, wdyt?

Ig-dolci:
Pessoal, acho melhor gastar energia quando esse PR estiver aberto. Tem coisas que irei mudar e as revisões ficarão obsoletas.

spyro/utils/physical_parameters.py
```diff
@@ -0,0 +1,236 @@
...
+    def __init__(self, fields=None):
+        self._fields = {}
+        if fields is None:
+            return
+        try:
+            items = fields.items()
+        except AttributeError:
+            items = fields
+        for name, value in items:
+            self.add(name, value)
```
ndf-ti, line 81: wdyt of forcing fields to be a map instead of trying to support both? The user can easily convert lists of tuple do a dictionary using `dict(list_of_tuples)`.

Also, maybe you can set the default of fields to a empty map like `__init__(self, field={})`.

With these changes you can simplify this code like this:

```Python

def __init__(self, field={}):
     self._fields = {}
     for name, value in fields.items():
          self.add(name, value)
```

spyro/utils/physical_parameters.py
```diff
@@ -0,0 +1,236 @@
+"""The named physical fields a wave equation is written in terms of."""
```
ndf-ti, line 1: can you add type hints to the methods arguments and returns?

spyro/utils/physical_parameters.py
```diff
@@ -0,0 +1,236 @@
...
+        """
+        return set(iterable)
+
+    def issubset(self, other):
+        """Return whether every parameter is also present in ``other``."""
+        return self <= other
+
+    def issuperset(self, other):
+        """Return whether ``other`` holds only parameters present here."""
+        return self >= other
```
ndf-ti, line 112: will this work? i didn't an override of `__le__` in this class. Same for `__ge__`

spyro/utils/physical_parameters.py
```diff
@@ -0,0 +1,236 @@
...
+        except KeyError:
+            raise KeyError(
+                f"'{key.value}' is not a physical parameter of this wave "
+                f"equation. Known parameters: {self._format_names()}.",
+            ) from None
+
+    def get(self, name, default=None):
+        """Return the field of ``name``, or ``default`` if it is not modelled."""
+        if name not in self:
+            return default
```
ndf-ti, line 129: i would also add a check to see if default is None:

```python
if name not in self and default is not None:
....
```
just so `get` doesn't return `None` without the user being aware.

spyro/solvers/elastic_wave/isotropic_wave.py
Ig-dolci: I think so. It looks like it is used to choose whether the wave equation is written as an explicit dependence on the Lame parameters or the velocity fields.

Olender (approved):
I really like these changes and don't have anything new to add

spyro/solvers/inversion.py
Olender: Acho que em outro PR podemos mover para outros arquivos e/ou tirar algumas coisas (tipo as classes do ROL e o caso sintético)

PR #384, opened 2026-08-21
Olender/i misplaced some plots

Olender:
This is a draft, since these plots are refactored versions of @ruansava 's commented out abc-related plots that I had accidentally deleted along with along other already refactored commented out code. Since I don't want to keep a lot of commented out code in main I am leaving these here as a draft until they can be tested with the newer abc features.

### DIFF
--- a/spyro/plots/general_plots.py
+++ b/spyro/plots/general_plots.py
@@ -230,7 +230,7 @@ def plot_shots(
     ----------
     wave : Wave
         Wave simulation object containing the shot record data in the
-        forward_solution_receivers attribute, along with timing and receiver information.
+        forward_solution_receivers attribute.
     show : bool, optional
         If True, display the plot interactively. Default is False.
     filename : str, optional
--- a/spyro/plots/plots_habc.py
+++ b/spyro/plots/plots_habc.py
@@ -8,6 +8,7 @@ import numpy as np
 
 from ..abc.lay_len import f_layer, loop_roots
 from .plot_helpers import _finalize_figure
+from ..utils.error_management import validate_numeric
 
 plt.rcParams.update({"font.family": "serif"})
 plt.rcParams["text.latex.preamble"] = r"\usepackage{bm} \usepackage{amsmath}"
@@ -203,8 +204,11 @@ def plot_function_layer_size(
 
 
 def plot_frequency_domain_receiver_responses(
-        wave, factor_xlim: float = 4.0, show: bool = False,
-    ):
+    wave,
+    frequency_limit_factor: float | int = 4.0,
+    output_folder: str | Path = "output/",
+    show: bool = False,
+):
     """Plot the frequency-domain receiver responses.
 
     Creates a multi-panel figure comparing the real FFT (RFFT) response of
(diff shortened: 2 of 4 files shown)


spyro/plots/general_plots.py
```diff
@@ -230,7 +230,7 @@ def plot_shots(
     ----------
     wave : Wave
```
ruansava, line 231: wave : `wave.Wave`
        An instance of the :class:`~spyro.solvers.wave.Wave`.
        Wave object containing the receiver data and simulation metadata.
        The following attributes are used:

        - ``dt``: simulation time step.
        - ``final_time``: final simulation time.
        - ``forward_solution_receivers``: receiver data from the simulation.
        - ``path_case_habc``: output directory for the generated figures.

spyro/plots/plots_habc.py
```diff
@@ -200,3 +201,249 @@ def plot_function_layer_size(
...
+):
+    """Plot the frequency-domain receiver responses.
+
+    Creates a multi-panel figure comparing the real FFT (RFFT) response of
+    each receiver between the computed and reference solutions. Vertical
+    lines indicate the source and reference frequencies.
+
+    Parameters
+    ----------
+    wave : object
```
ruansava, line 220: wave : `wave.Wave`
        An instance of the :class:`~spyro.solvers.wave.Wave`.
        Wave object containing the simulation results. It must provide the
        following attributes:
        - ``receivers_out_fft`` : ndarray
            RFFT of the computed receiver data. The first dimension
            corresponds to frequency and the second to receivers.
        - ``receivers_ref_fft`` : ndarray
            RFFT of the reference receiver data, with the same shape as
            ``receivers_out_fft``.
        - ``dt`` : float
            Time step used in the simulation, in seconds.
        - ``frequency`` : float
            Source frequency in Hz.
        - ``freq_ref`` : float
            Reference frequency in Hz.
        - ``number_of_receivers`` : int
            Number of receivers.

spyro/plots/plots_habc_damping_factor_curve.py
ruansava, line 1: It should be in plots_habc.py

ruansava (changes requested):

spyro/plots/plots_habc_damping_factor_curve.py
ruansava, line 1: The plot should generate something like this

Image

spyro/plots/general_plots.py
ruansava: Why are you using double (``) instead single (')?

spyro/plots/receiver_plots.py
ruansava, line 375: Please, modify this line so the legend can be shown without overlapping the plot

`receiver_axes[0].legend(ncol=2, fontsize=6, loc='best')`

ruansava (changes requested):

spyro/plots/plots_habc.py
```diff
@@ -200,3 +201,249 @@ def plot_function_layer_size(
...
+
+def plot_frequency_domain_receiver_responses(
+    wave,
+    frequency_limit_factor: float | int = 4.0,
+    output_folder: str | Path = "output/",
+    show: bool = False,
+):
+    """Plot the frequency-domain receiver responses.
+
+    Creates a multi-panel figure comparing the real FFT (RFFT) response of
```
ruansava, line 214: It is not the real part of the FFT, it is the magnitude of FFT. I prefer frequency response

"Creates a multi-panel figure comparing the frequency response magnitude
    obtained via FFT of each receiver between the computed and reference solutions.
    Vertical lines indicate the source and reference frequencies."

ruansava (changes requested):

PR #385, opened 2026-08-21
Automated adjoint for the isotropic elastic wave

Ig-dolci:
Extends the automated adjoint (pyadjoint, via `firedrake.adjoint`) from the
acoustic solver to the isotropic elastic one, with selectable material
controls.

### What this adds

**Multiple controls.** `AutomatedAdjoint` held exactly one control, fixed to
`wave.c`. It now holds an ordered list, builds the reduced functional from a
list of `Control`s, and returns one derivative per control. A single control
still returns a bare `Function`, following pyadjoint's own convention, so the
acoustic API is unchanged.

**Selectable parameters.** `enable_automated_adjoint(control_parameters=...)`
names the parameters to differentiate with respect to. They are resolved
before any adjoint state exists, so an invalid selection fails early rather
than half-way through. Passing nothing selects everything the equation
offers.

**Change of variables.** An isotropic elastic medium has five named material
parameters and three degrees of freedom. All five are read regardless: the
variational form is written in density and the Lame parameters, while the
absorbing boundary conditions and the stable timestep estimate are written in
the two wave speeds.

What the choice decides is which three carry the data:

- declaring `{density, lambda, mu}` computes
  `p_wave_velocity = sqrt((lambda + 2*mu)/density)` and
  `s_wave_velocity = sqrt(mu/density)`;
- declaring `{density, p_wave_velocity, s_wave_velocity}` computes
  `mu = density*s_wave_velocity**2` and
  `lambda = density*p_wave_velocity**2 - 2*mu`.

The three that carry the data are `Function` fields; the other two are UFL
expressions of them. This is not cosmetic: the variational form embeds those
expressions, so the assembled problem depends symbolically on whichever
fields are the source, and the chain rule reaches only those. Differentiating
with respect to `p_wave_velocity` requires it to be a field, because a
computed expression has no degrees of freedom to perturb — and when the Lame
parameters are the fields, `p_wave_velocity` does not appear in the form at
all.

`Wave.set_physical_parameterization()` moves the source from one set to the
other, so an inversion is not tied to the parameters its model happens to be
stored in. Which set is used changes the conditioning of the problem and the
cross-talk between parameters, so it is worth choosing deliberately. It is
chosen once, before the adjoint is enabled.

**A fix in `get_real_shot_record`.** Elastic records are
`(n_timesteps, n_receivers, dimension)`. Rank was compared against a
constant, so one elastic shot was indistinguishable from several acoustic
ones. It now compares against the rank the wave type implies. This only
manifests with multiple shots, which is why the parallel test covers it.

### Design notes

`AutomatedAdjoint` gets everything it needs about the forward problem from
the tape that the forward solve recorded — which is what it differentiates.
What it does not have is a reference to the wave equation, or any branch on
wave type, parameterisation or medium: it receives the controls, already
resolved, and the ensemble to sum over, and works from the recording.

Resolving names to fields is `PhysicalParameters.select()`, since it is
entirely about the container's contents: a name is selectable exactly when it
is carried by a field rather than computed from the others.

Elastic material parameters are now built once rather than on every forward
solve. The acoustic solver already behaved this way. Rebuilding replaced the
objects the adjoint is posed with respect to, and would reset an inversion's
current iterate to the model's initial values. Replacing the model clears the
parameters, which is what allows them to be built again.

### Tests

- `tests/on_one_core/test_elastic_auto_adj_2d.py` — Taylor tests over four
  cases: both parameterizations with every parameter as a control, a
  single-parameter subset, and a two-parameter subset after a change of
  variables. Convergence rate above 1.9 in all four.
- `tests/parallel/test_elastic_sources_parallel.py` — the ensemble-summed
  multi-control gradient under source parallelism, checking each control's
  gradient is identical across ensemble members. Added to the CI job that
  runs the parallel inversion tests.
- Unit tests for control normalization, shot-record ranks, and the rejection
  of parameters the equation carries as expressions.

### DIFF
--- a/spyro/solvers/automatic_differentiation_solver.py
+++ b/spyro/solvers/automatic_differentiation_solver.py
@@ -14,9 +14,9 @@ def _as_list(value: object) -> list:
 
     Parameters
     ----------
-    value : object, mapping, list, tuple, or None
-        Value to normalize. Mappings contribute their values and ``None``
-        produces an empty list.
+    value : object, mapping, PhysicalParameters, list, tuple, or None
+        Value to normalize. Anything keyed by name contributes its values,
+        and ``None`` produces an empty list.
 
     Returns
     -------
@@ -25,7 +25,7 @@ def _as_list(value: object) -> list:
     """
     if value is None:
         return []
-    if isinstance(value, Mapping) or hasattr(value, "items"):
+    if isinstance(value, (Mapping, PhysicalParameters)):
         return list(value.values())
     if isinstance(value, (list, tuple)):
         return list(value)
@@ -70,11 +70,18 @@ class AutomatedAdjoint:
     .. code-block:: python
 
         wave.enable_automated_adjoint(control_parameters=parameters)
-        with wave.automated_adjoint.fresh_tape():
-            wave.forward_solve()          # forward run recorded on the tape
+        wave.forward_solve()          # the time integrator starts recording
+        dJ = wave.gradient_solve()    # one derivative per selected parameter
+        wave.automated_adjoint.clear_tape()
+
+    ``gradient_solve`` builds the reduced functional from the recorded tape
+    on its first call. Building it here is only needed to hold on to it, or
+    to run a Taylor test against it:
+
+    .. code-block:: python
+
         wave.automated_adjoint.create_reduced_functional(wave.functional_value)
-        dJ = wave.automated_adjoint.compute_gradient()
-        rate = wave.automated_adjoint.verify_gradient(wave.c)  # Taylor test
+        rate = wave.automated_adjoint.verify_gradient(wave.c)
 
     Parameters
     ----------
@@ -106,33 +113,17 @@ pyadjoint.ReducedFunctional or None
     """
 
     def __init__(self, ensemble: object, controls: object = None) -> None:
-        self.controls = []
-        self.control_parameter_names = []
         self.ensemble = ensemble
         self.reduced_functional = None
         self._tape = None
-        self._set_controls(controls)
-
-    def _set_controls(self, controls: object) -> None:
-        """Store labeled or unlabeled control fields as ordered lists.
-
-        Parameters
-        ----------
-        controls : object or iterable
-            Control fields, optionally exposed through an ``items`` method.
-
-        Returns
-        -------
-        None
-        """
-        try:
-            control_items = list(controls.items())
-        except AttributeError:
+        if isinstance(controls, (Mapping, PhysicalParameters)):
+            # Keyed by material parameter: keep the labels, so the
+            # derivatives can be handed back under the same names.
+            self.control_parameter_names = list(controls)
+            self.controls = list(controls.values())
+        else:
             self.controls = _as_list(controls)
             self.control_parameter_names = [None] * len(self.controls)
-        else:
-            self.control_parameter_names = [name for name, _ in control_items]
-            self.controls = [value for _, value in control_items]
 
     @contextmanager
     def fresh_tape(self):
--- a/spyro/solvers/automatic_differentiation_solver.py
+++ b/spyro/solvers/automatic_differentiation_solver.py
@@ -277,7 +277,11 @@ pyadjoint.ReducedFunctional or None
         """
         if not self.controls:
             raise ValueError("At least one control is required.")
-        control = [fire_ad.Control(value) for value in self.controls]
+        controls = [fire_ad.Control(value) for value in self.controls]
+        # pyadjoint mirrors the shape it is given: a bare control comes back
+        # as a bare derivative, a list as a list. Handing it a one-item list
+        # would make every single-control caller unwrap by hand.
+        control = controls[0] if len(controls) == 1 else controls
 
         self.reduced_functional = fire_ad.EnsembleReducedFunctional(
             functional,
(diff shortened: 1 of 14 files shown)


spyro/solvers/elastic_wave/isotropic_wave.py
```diff
@@ -109,154 +114,255 @@ def initialize_model_parameters_from_object(self, synthetic_data_dict: dict):
...
+            objects, or UFL expressions. The dictionary is only read from.
 
         Returns
         -------
         None
             The method assigns ``rho``, ``lmbda``, ``mu``, ``c``, ``c_s``, and
-            the active control parameterization on ``self``.
+            the active physical parameterization on ``self``.
         """
-        def material_parameter(value):
```
Ig-dolci, line 120: `material_parameter` não foi removido. Foi dividido porque fazia duas coisas: lia o valor do dicionário normalizando escalar → `Constant`, e construía o `Function` quando já havia malha. As duas
continuam existindo, agora separadas:

- a leitura ficou em `declared`, logo abaixo;
- a construção do campo foi para `as_function`, em
  `_build_physical_parameterization`.

spyro/solvers/elastic_wave/elastic_wave.py
```diff
@@ -53,6 +53,49 @@ def initialize_model_parameters_from_object(self, synthetic_data_dict):
...
+        Initializing the material properties already picks a set, by reading
+        whichever one the model input declares. This method is public because
+        that first choice is worth revising: an inversion is not obliged to
+        invert in the parameters its model happens to be stored in, and which
+        set it uses changes the conditioning of the problem and the cross-talk
+        between parameters. Exposing the change of variables keeps it in the
+        solver, where it is the same one initialization performs, rather than
+        leaving callers to convert their model input by hand and get a factor
+        or a sign wrong.
+
```
Ig-dolci:
```suggestion
```

spyro/solvers/automatic_differentiation_solver.py
```diff
@@ -43,19 +69,28 @@ class AutomatedAdjoint:
     ----------------
     .. code-block:: python
 
-        wave.enable_automated_adjoint()   # builds AutomatedAdjoint(wave.comm)
-        with wave.automated_adjoint.fresh_tape():
-            wave.forward_solve()          # forward run recorded on the tape
+        wave.enable_automated_adjoint(control_parameters=parameters)
```
Olender, line 72: Parei aqui

spyro/solvers/automatic_differentiation_solver.py
```diff
@@ -43,19 +69,28 @@ class AutomatedAdjoint:
...
+        wave.forward_solve()          # the time integrator starts recording
+        dJ = wave.gradient_solve()    # one derivative per selected parameter
+        wave.automated_adjoint.clear_tape()
```
Olender, line 75: Why the clear tape instead of with fresh tape as before?

spyro/utils/utils.py
```diff
@@ -645,7 +645,8 @@ def get_real_shot_record(wave):
     """Get the real shot record for the active sources.
 
     The returned object is typically an array with shape
-    ``(n_timesteps, n_receivers)`` for a single active shot.
+    ``(n_timesteps, n_receivers)`` for a scalar wave or
+    ``(n_timesteps, n_receivers, dimension)`` for a single elastic shot.
```
Olender, line 649: parei aqui

tests/parallel/test_elastic_sources_parallel.py
```diff
@@ -0,0 +1,160 @@
...
+        convergence_rate = wave.automated_adjoint.verify_gradient(
+            wave.automated_adjoint.controls,
+            direction=directions,
+            dJdm=gradients,
+        )
+        assert convergence_rate > 1.9, (
+            "Parallel elastic Taylor convergence rate %.4f < 1.90."
+            % convergence_rate
+        )
+    finally:
```
Olender, line 155: If an error occurs during try and we, therefore, don't pass an assert, these tests will pass right? If so, we need to make sure they fail

Olender (changes requested):

Olender:
Gostei bastante das mudanças como um todo. Meu principal comentário é em relação aos testes. Acho que mesmo falhando num try, se não apssar por um assert o teste 'passa'

spyro/solvers/automatic_differentiation_solver.py
Ig-dolci: The example was wrong. Recording starts in the time integrator: `wave_propagator` calls `automated_adjoint.start_recording()` once `enable_automated_adjoint` has set the adjoint type. So `forward_solve` already manages the tape, and wrapping it in `fresh_tape` is unnecessary because `fresh_tape` is not called anywhere in `spyro/` or `tests/`.

The example now shows what every test actually does: enable, solve, take the
gradient, clear the tape.

Clear tape is a good practice after finalising a single gradient computation; that is what this PR supports. In future, I will update this code block for the FWI case.

tests/parallel/test_elastic_sources_parallel.py
Ig-dolci: No, the test fails. Actually, this is the safest approach because the finally block in Python runs cleanup actions regardless of whether an exception is raised or handled. That is why `finally` is used. The tape has to be cleared on the failing path too. Leaving a tape behind would make the *next* test in that process record on top of it 😱 (Haha).

Olender: I might be wrong here, I just want to check before merging

Ig-dolci: You can verify it in a simple test:
```
try:
    raise RuntimeError("boom")
    assert False
finally:
    print("cleanup ran")
```

Olender (approved):

Issue #326, opened 2026-07-07
Create SpyroReducedFunctional


Ig-dolci:
`SpyroReducedFunctional` will be pretty interesting to use the optimisation solvers available in pyadjoint. `SpyroReducedFunctional` should support automated and implemented adjoints.

Issue #315, opened 2026-06-29
VertexOnlyMesh receiver sampling fails under spatial (mesh) parallelism


Ig-dolci:
## Summary

Sampling receivers through a vertex-only mesh (`use_vertex_only_mesh = True`) is incompatible with decomposing a shot's mesh across more than one MPI rank. It works with **one core per shot** (serial or ensemble/shot parallelism). This is a VOM receiver-handling problem and is **independent of the automated adjoint**. The automated adjoint merely turns the flag on, so its parallel test is where it surfaced.

## Where it shows up

- `tests/parallel/test_gradient_auto_adjoint.py::test_gradient_auto_adjoint_parallel`
  with mesh parallelism (e.g. `mpiexec -n 4` → 2 shots × 2 spatial cores).
  It is currently pinned to 1 core per shot (`mpiexec -n 2`) as a workaround.
- More generally: any forward solve with `use_vertex_only_mesh = True` and `>1`
  core per shot (no adjoint required).

## Error

```
pyop2.exceptions.DataValueError: Invalid data: expected 5 values, got 10!
pyop2.exceptions.DataValueError: Invalid data: expected 6 values, got 10!
```

(`10` = number of receivers; `5`/`6` = receivers owned by each spatial rank.)

## Root cause

A vertex-only mesh distributes the receivers. Then, each rank only owns the ones inside its mesh partition, but the receiver code assumes every rank holds a **global-width** array. Two places break:

1. **Receiver finalization** (runs for any VOM forward solve), `helpers.fill` +
   `utils.communicate`: `fill` writes over the global receiver count on a
   local-width array, and `communicate` does an `MPI.MAX` `Allreduce` that needs
   equal, global-width arrays on every rank → `IndexError` / shape mismatch.
2. **Per-timestep misfit** (automated-adjoint path),
   `spyro/solvers/time_integration_central_difference.py`:
   ```python
   real_shot = fire.Function(usol_recv[-1].function_space(), val=real_shot_record[step])
   ```
   puts the global record (all receivers) into the local VOM space (5/6 DOFs) →
   the `DataValueError` above.

Olender:
We need to fix this before Issue #319

Olender:
Do you know if this is a spyro issue or a Firedrake issue?

Ig-dolci:
It is a Spyro issue. It is the way Spyro is returning the real shot array for a Firedrake Function. I believe this will not be a big headache to fix.

[...]

Issue #92, opened 2023-09-21
Demos for automated adjoint


Ig-dolci:

Olender:
There are no demos anymore. Should we just close this issue and leave the automated adjoint for a jupyter notebook?

PR #391, opened 2026-08-28
Fwi support for automated adjoints

Ig-dolci:
`FullWaveformInversion` could only be run with the implemented adjoint, which exists only for acoustic media. This PR seeks to run with the automated adjoint too for acoustic and isotropic elastic cases. 

### What changes for a user

Nothing for acoustic and implemented adjoints

If the user wants to use automated adjoint. Then, one argument to `run_fwi`:

```python
fwi.run_fwi(
    adjoint_type=AdjointType.AUTOMATED_ADJOINT,
    adjoint_options={"checkpointing": True, "snapshots": 10},
    vmin=[1.0, 1.5, 0.5], vmax=[3.0, 4.0, 2.5], maxiter=20,
)
```
for elastic.

Everything before that call is unchanged. The default path — no `adjoint_type`
— still runs L-BFGS-B on the implemented adjoint, exactly as before.

### How it works

The two adjoints reach the optimiser by different routes, and that is the
substance of the change:

| | implemented adjoint | automated adjoint |
|---|---|---|
| forward solves | one per optimizer iterate | **one**, recorded on a pyadjoint tape |
| optimizer | `scipy.optimize.minimize`, L-BFGS-B | PETSc TAO (`blmvm`) |
| control | flat vector of degrees of freedom | the `Function`s themselves |
| bounds | one pair per degree of freedom | one pair per control |

### Multi-parameter controls

An acoustic inversion has one control, its velocity model, and
`control_parameters` still returns that single `Function`. An isotropic elastic
one is inverted for any subset of whichever set the equation is written in —
density and the two wave speeds, or density and the Lamé parameters — and those
come back as a list. Everything the optimisers see follows from that: bounds
take one entry per control, derivatives come back one per control keyed by
their parameter, and `run_fwi` returns the controls it converged on.

Which parameters are inverted for is
`adjoint_options={"control_parameters": {...}}`; density can be held fixed
while the two Lamé parameters are fitted.

### New public surface

- `set_real_model(material)` — the true model of a synthetic experiment, for a
  medium of any kind. `set_real_velocity_model` is acoustic-only and now emits
  a `DeprecationWarning`; the three forms its callers use (constant,
  conditional, expression string) produce byte-identical observed data through
  the new one.
- `run_fwi(adjoint_type=..., adjoint_options=...)`.

### `spyro/tools/optimization.py`

New module holding what belongs to the optimizer rather than to the inversion:
`minimize_with_tao` and `tao_bounds`. All PETSc/pyadjoint solver imports leave
`inversion.py`, including the recovery of the last iterate when TAO stops at
its iteration limit — which is the normal way FWI is run, and is warned about
rather than raised.

### Tests

| file | ranks | what it covers |
|---|---|---|
| `tests/on_one_core/test_fwi_automated_adjoint.py` | 1 | acoustic and elastic inversions: result shape, bounds honoured, every control moved, functional decreased, optimum written back |
| `tests/parallel/test_fwi_auto_adjoint.py` | 2 | the same two media under shot parallelism, plus that both ensemble members converge on the same model and the same functional history |

The parallel pair is what catches a TAO built on the wrong communicator.


Ig-dolci:
Demos are coming soon. We already have issue #92 for this.

spyro/solvers/inversion.py
```diff
@@ -425,83 +483,112 @@ def control_parameters(self):
...
+           medium with two parameterizations declares both sets, so this last
+           answer is only exact where there is one set to choose from.
 
-        Cases 2 and 3 are what lets ``set_guess_control`` work on a fresh
-        inversion: it is handed a bare ``Function`` that carries no indication
-        of which parameter it belongs to, and the answer has to come from
-        somewhere.
+        Case 2 is what lets ``set_guess_control`` work on a fresh inversion:
+        it is handed a bare value that carries no indication of which
+        parameter it belongs to, and the answer has to come from somewhere.
```
Olender, line 568: Gostei dessa explicação

spyro/solvers/inversion.py
```diff
@@ -1309,82 +1622,332 @@ def return_functional_and_gradient(self, c):
...
+        spyro.tools.optimization.minimize_with_tao :
+            The optimizer this hands the reduced functional to, and what it
+            does when TAO stops short of converging.
+        """
+        # Imported here, not at module load: the optimizer is built on
+        # pyadjoint's TAO support, and only this path needs it. A driver that
+        # never asks for the automated adjoint must not be made to depend on
+        # it -- ``import spyro`` reaches this module, so anything unavailable
+        # up there fails every test that touches spyro, whatever it tests.
+        from ..tools.optimization import minimize_with_tao, tao_bounds
```
Olender, line 1858: Is this made just to avoid a circular import that would happen with a module level import?

spyro/utils/physical_parameters.py
```diff
@@ -1,11 +1,49 @@
 """The named physical fields a wave equation is written in terms of."""
 
-from collections.abc import Set
+from collections.abc import Mapping, Set
 from enum import Enum
 
 import firedrake as fire
 
 
+def as_list(value):
```
Olender, line 9: Parei aqui

tests/on_one_core/test_fwi_automated_adjoint.py
```diff
@@ -0,0 +1,218 @@
...
+        "type": "object",
+        **material,
+        "real_velocity_file": None,
+    }
+    return dictionary
+
+
+@pytest.mark.newer_firedrake
+def test_fwi_automated_adjoint(tmp_path, monkeypatch):
+    """Invert an acoustic velocity model from data a faster one generated.
```
Olender, line 119:
```suggestion
    """Invert an acoustic velocity model.
```

tests/on_one_core/test_fwi_automated_adjoint.py
```diff
@@ -0,0 +1,218 @@
...
+    # The monitor logged the iterates on top of the recorded starting point,
+    # and the run brought the functional down.
+    assert len(fwi.functional_history) > 1
+    assert fwi.functional_history[-1] < fwi.functional_history[0]
+    assert fwi.functional == fwi.functional_history[-1]
+
+    # The optimum is written back into the driver and the solver alike.
+    assert np.allclose(fwi.control_parameter_result.dat.data_ro, values)
+    assert np.allclose(fwi.wave.c.dat.data_ro, values)
+    assert (tmp_path / "result.npy").exists()
```
Olender, line 164: Would you mind adding a demo similar to this in a future PR?

spyro/solvers/inversion.py
```diff
@@ -1309,82 +1622,332 @@ def return_functional_and_gradient(self, c):
...
+            self.set_guess_control(self.control_parameter_result)
+
+        fire.VTKFile("control_end.pvd").write(
+            *as_list(self.control_parameter_result),
+        )
+
+        np.save("result", self._flatten_control(self.control_parameter_result))
+        return result
+
+    def _run_fwi_tao(self, parameters, tao_options=None):
```
Olender, line 1816: Do you think it's possible to add the TAO solver to the implemented adjoint in the future in another PR?

tests/parallel/test_fwi_auto_adjoint.py
```diff
@@ -0,0 +1,241 @@
...
+
+    history_from_root = ensemble_comm.bcast(fwi.functional_history, root=0)
+    assert np.allclose(fwi.functional_history, history_from_root), (
+        "Ensemble members disagree on the functional they minimized."
+    )
+    assert fwi.functional_history[-1] < fwi.functional_history[0]
+
+
+@pytest.mark.newer_firedrake
+@pytest.mark.parallel(2)
```
Olender, line 152: O que acha de mudar esse teste pra 4 cores (sem modificar o uso das 2 fontes) para garantirmos o funcionamento do comm espacial também?

Olender (changes requested):
What do you think of changing the parallel fwi test to run in 4 cores with 2 sources?

tests/on_one_core/test_fwi_automated_adjoint.py
Ig-dolci: I am currently working on a demo that is coming very soon. We already have an issue #92  for this.

spyro/solvers/inversion.py
Ig-dolci: Yes. I think the best approach is to use the pyadjoint TAO code instead of implementing TAO for the implemented adjoint in Spyro, which will require some time. For this reason, I opened issue #326. With a Spyro-reduced functional, we can easily use the TAO optimisation solver from pyadjoint. I think someone mentioned this approach to you at the Firedrake workshop.

tests/parallel/test_fwi_auto_adjoint.py
Ig-dolci: Acho que temos um bug pra paralisação espacial com o VOM. Não encontrei a issue. Isso foi corrigido?

spyro/solvers/inversion.py
Ig-dolci: No. This is to avoid import pyadjoint's objects when it is not using automated adjoints. Also, I think I had an error when importing in a module level, nothing related to circular import. Perhaps firedrake version, I am not sure.

tests/parallel/test_fwi_auto_adjoint.py
Olender: Ahhh verdade, achei que isso já tinha sido resolvido. É o Issue #315

Olender (approved):

PR #399, opened 2026-09-03
Move Read_options and Read_outputs to pydantic

ndf-ti:
This PR moves the Read_options and Read_outputs to use pydantic so we can automate a lot of the validations being done as pydantic can automaticaly do typechecking at runtime. While the model classes are not ideal, i think this is a good first step to migrate away from using dictionaries.

### DIFF
--- a/spyro/io/model_parameters.py
+++ b/spyro/io/model_parameters.py
@@ -13,10 +13,9 @@ from .. import utils
 from .. import meshing
 
 
-class Model_parameters(
-    Read_options, Read_boundary_layer, Read_time_axis, Read_outputs, VelocityModelFileIO
-):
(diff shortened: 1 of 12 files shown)


spyro/io/model_parameters.py
```diff
@@ -557,6 +557,20 @@ def domain_dimensions(self):
 
         return domain_dim
 
+    # This is a temporary approach to no break the current API. This way it is still possible to call
```
Olender, line 560: Good idea! Is there a way to add a deprecated flag if someone does this?

Olender:
I think this PR is ready as soon as the tests pass

Olender (approved):

PR #401, opened 2026-09-09
Fix/adjoint memory

Ig-dolci:


spyro/solvers/backward_time_integration.py
```diff
@@ -88,16 +98,29 @@ def backward_wave_propagator(wave: Wave, dt: float = None) -> fire.Function:
...
+                    if len(forward_solution) > 0:
+                        forward_solution.peek_into(forward_field, 0)
+                    else:
+                        forward_field.assign(0.0)
                 else:
-                    forward_field.assign(0.0)
+                    forward_solution.pop()
+                    if len(forward_solution) > 0:
+                        forward_field.assign(forward_solution[-1])
+                    else:
```
ndf-ti, line 114: maybe we can move the forward assign out of the if block like:

```python
if using_store:
    forward_solution.pop_into(forward_field)
    if len(forward_solution) > 0:
          forward_solution.peek_into(forward_field, 0)
else:
    forward_solution.pop()
    if len(forward_solution) > 0:
          forward_field.assign(forward_solution[-1])

if len(forward_solution) == 0:
    forward_field.assign(0.0)
```

ndf-ti (approved):

PR #403, opened 2026-09-10
Refactor acoustic and elastic weak-form builders

j-salesps:
To build the coupled fluid-solid solver, the acoustic and elastic weak-form constructors needed to be adapted so that only the weak form (matrices) is built, without invoking their respective solvers. Both were therefore split into two functions each, one that builds the form and one that assembles the solver, so that none of the existing simulation modes are broken (acoustic-only, elastic-only, and the coupled acoustic-elastic solver).

**Changes:**

- Split `build_acoustic_form` / `construct_solver_or_matrix_no_pml` and `build_elastic_form` / `isotropic_elastic_no_pml` into separate form-building and solver-construction functions;
- Added support for specifying the acoustic wave speed via the fluid bulk modulus (`K`) and density (`rho_fluid`) instead of `c` directly, since these parameters will be needed for the future interface-continuity analysis in the coupled solver. `c` and `K`/`rho_fluid` are mutually exclusive; passing both, or only one of `K`/`rho_fluid`, raises a clear `ValueError`;
- General formatting and docstring cleanup. Behavior is otherwise unchanged.

 **Follow-up:**

A future PR will rename the solid density attribute from `rho` to `rho_solid`, for clarity and consistency with the fluid-side `rho_fluid` naming.

### DIFF
--- a/spyro/solvers/acoustic_solver_construction_no_pml.py
+++ b/spyro/solvers/acoustic_solver_construction_no_pml.py
@@ -1,21 +1,54 @@
-"""Constructs Firedrake solver for the acoustic wave with typical BCs, 
-NRBCs or HABCs."""
+"""Constructs Firedrake solver for the acoustic wave with BCs, NRBCs, HABCs."""
 
 import firedrake as fire
 from firedrake import ds, dx, dot, grad, sqrt
 from ..utils.typing import AbsorbingBCsType
 
 
+def velocity_fluid(c=None, K=None, rho_fluid=None, default=None):
+    """Resolve the acoustic wave velocity.
+
+    Uses either an explicity value or the fluid bulk modulus and density.
+
+    Parameters
+    ----------
+    c : `firedrake.Constant` or UFL expression, optional
+        Fluid wave speed. Mutually exclusive with `K`/`rho_fluid`.
+    K : `firedrake.Constant` or `firedrake.Function`, optional
+        Fluid bulk modulus. Must be given together with `rho_fluid`.
+    rho_fluid : `firedrake.Constant` or `firedrake.Function`, optional
+        Fluid density. Must be given together with `K`.
+    default : `firedrake.Constant` or UFL expression, optional.
+        Value to return when neither `c` nor `K`/`rho_fluid` are given.
+
+    Returns
+    -------
+    c : UFL expression
+        The resolved wave velocity.
+
+    Raises
+    ------
+    ValueError
+        If `c` is given together with `K` or `rho_flui`, or if
+        only one of `K`/`rho_fluid` is given.
+    """
+    if c is not None and (K is not None or rho_fluid is not None):
+        raise ValueError(
+            "Ambiguous formulation. Insert 'bulk_modulus' and 'density_fluid' "
+            "or just 'velocity_fluid'."
+        )
+    if c is None:
+        if K is not None and rho_fluid is not None:
+            c = sqrt(K / rho_fluid)
+        elif K is not None or rho_fluid is not None:
+            raise ValueError("Insert both values 'bulk_modulus' and 'density_fluid'.")
+        else:
+            c = default
+    return c
+
+
 def build_acoustic_form(
-    wave,
-    u_trial,
-    v_test,
-    u_n,
-    u_nm1,
-    quad_rule,
-    c=None,
-    K=None,
-    rho_fluid=None
+    wave, u_trial, v_test, u_n, u_nm1, quad_rule, c=None, K=None, rho_fluid=None
 ):
     """Build the weak form of the acoustic wave equation for one time step.
 
@@ -33,12 +66,9 @@ def build_acoustic_form(
         Pressure field at the previous time step.
     quad_rule : dict
         Quadrature rule for volume integration.
-    c : `firedrake.Constant` or UFL expression, optional
-        Fluid wave speed. Mutually exclusive with `K`/`rho_fluid`.
-    K : `firedrake.Constant` or `firedrake.Function`, optional
-        Fluid bulk modulus. Must be given together with `rho_fluid`.
-    rho_fluid : `firedrake.Constant` or `firedrake.Function`, optional
-        Fluid density. Must be given together with `K`.
+    c, K, rho_fluid : optional
+        Passed through to `velocity_fluid` to determine the fluid wave
+        velocity.
 
     Returns
     -------
@@ -51,20 +81,7 @@ def build_acoustic_form(
         If `c` is given together with `K` or `rho_fluid` (ambiguous), or if
         only one of `K`/`rho_fluid` is given (incomplete).
     """
-
-    if c is not None and (K is not None or rho_fluid is not None):
-        raise ValueError(
-            "Ambiguous formulation. Insert 'bulk_modulus' and 'density_fluid' "
-            "or just 'c'."
-        )
-    if c is None:
-        if K is not None and rho_fluid is not None:
-            c = sqrt(K / rho_fluid)
-        elif K is not None or rho_fluid is not None:
-            raise ValueError("Insert both values 'bulk_modulus' and 'density_fluid'.")
-        else:
-            c = wave.c
-
+    c = velocity_fluid(c, K, rho_fluid, default=wave.c)
     dt = wave.dt
 
     m1 = (
@@ -123,15 +140,15 @@ def build_acoustic_form(
 
 
 def construct_solver_or_matrix_no_pml(wave):
-    """Build the Firedrake solver for the acoustic wave, without PML,
-    with typical BCs, NRBCs or HABCs.
+    """Build the Firedrake solver for the acoustic wave, without PML.
+
+    Handles typical BCs, NRBCs or HABCs.
 
     Parameters
     ----------
     wave : `acoustic_wave.AcousticWave`
         An instance of the :class:`~spyro.solvers.acoustic_wave.AcousticWave`.
     """
-
     V = wave.function_space
     quad_rule = wave.quadrature_rule
 
--- a/spyro/solvers/acoustic_solver_construction_no_pml.py
+++ b/spyro/solvers/acoustic_solver_construction_no_pml.py
@@ -1,20 +1,137 @@
-"""Constructs Firedrake solver for the acosutic wave with typical BCs, NRBCs or HABCs."""
+"""Constructs Firedrake solver for the acoustic wave with typical BCs, 
+NRBCs or HABCs."""
 
 import firedrake as fire
-from firedrake import ds, dx, dot, grad
+from firedrake import ds, dx, dot, grad, sqrt
 from ..utils.typing import AbsorbingBCsType
 
 
-def construct_solver_or_matrix_no_pml(wave):
-    """Builds solver operators for wave propagator with typical BCs, NRBCs or HABCs.
+def build_acoustic_form(
+    wave,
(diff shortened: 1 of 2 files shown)


spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -1,55 +1,89 @@
...
+            "Ambiguous formulation. Insert 'bulk_modulus' and 'density_fluid' "
+            "or just 'c'."
+        )
+    if c is None:
+        if K is not None and rho_fluid is not None:
+            c = sqrt(K / rho_fluid)
+        elif K is not None or rho_fluid is not None:
+            raise ValueError("Insert both values 'bulk_modulus' and 'density_fluid'.")
+        else:
+            c = wave.c
```
ndf-ti: i think that this function could just receive a valid c right away. Maybe you can move this calculation of c using K and rho_fluid to another function, then the user can use that function if he want.

Olender:
I noticed the linting check failed. If you want you could run `make lint` in your terminal to check which lines in your code will fail the check

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -58,40 +109,75 @@ def construct_solver_or_matrix_no_pml(wave):
...
+    u_n = fire.Function(V, name="pressure")
+    u_np1 = fire.Function(V, name="pressure t+dt")
+    wave.u_nm1 = u_nm1
+    wave.u_n = u_n
+    wave.u_np1 = u_np1
+
+    wave.current_time = 0.0
+    dt = wave.dt
+
+    form = build_acoustic_form(wave, u, v, u_n, u_nm1, quad_rule)
```
Olender, line 169: Why do you need to pass u_n, u_nm1, and the quadrature rule if they are already attributes of the wave object?

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -58,40 +109,75 @@ def construct_solver_or_matrix_no_pml(wave):
...
+
+    u_nm1 = fire.Function(V, name="pressure t-dt")
+    u_n = fire.Function(V, name="pressure")
+    u_np1 = fire.Function(V, name="pressure t+dt")
+    wave.u_nm1 = u_nm1
+    wave.u_n = u_n
+    wave.u_np1 = u_np1
+
+    wave.current_time = 0.0
+    dt = wave.dt
```
Olender, line 167:
```suggestion
```

Olender: This is where the linting check fails

Olender:
> To build the coupled fluid-solid solver, the acoustic and elastic weak-form constructors needed to be adapted so that only the weak form (matrices) is built, without invoking their respective solvers. Both were therefore split into two functions each, one that builds the form and one that assembles the solver, so that none of the existing simulation modes are broken (acoustic-only, elastic-only, and the coupled acoustic-elastic solver).

Good idea! I really like these piece-meal PRs

Olender (changes requested):
Just two changes:
- linting
- @ndf-ti 's comment. We should really be careful with not redefining the physical parameters unnecessarily. This should be done in a function outside of this. We should talk to @Ig-dolci, who recently added the physical parameters, just to be sure

spyro/solvers/elastic_wave/forms.py
```diff
@@ -1,39 +1,97 @@
...
-    V = wave.function_space
-    quad_rule = wave.quadrature_rule
+from .local_abc import local_abc_form
 
-    u = TrialFunction(V)
-    v = TestFunction(V)
 
-    u_nm1 = wave.u_nm1
-    u_n = wave.u_n
+def build_elastic_form(wave, u_trial, v_test, u_n, u_nm1, quad_rule):
```
ruansava, line 25: @Olender 
`u_trial` and `v_test` seem like suitable names for these types of variables. Shall we adopt them for the future?

spyro/solvers/elastic_wave/forms.py
```diff
@@ -1,39 +1,97 @@
...
-        F_s += dot(b, v)*dx(**quad_rule)
+        F_s += dot(b, v_test) * dx(**quad_rule)
 
     F_t = local_abc_form(wave)
 
-    F = F_m + F_k - F_s - F_t
+    return F_m + F_k - F_s - F_t
+
+
+def isotropic_elastic_without_pml(wave):
```
ruansava, line 74: I am not sure if forcing a name like "without_pml" is a good idea.
It should be a general solver; if any specific feature is included, it will become part of the new name. In fact, I will not implement PML in the elastic solver. because it is very expensive

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -1,55 +1,106 @@
-"""Constructs Firedrake solver for the acosutic wave with typical BCs, NRBCs or HABCs."""
+"""Constructs Firedrake solver for the acoustic wave with BCs, NRBCs, HABCs."""
```
ruansava, line 1: I defend the use of the typical BCs (Neumann or Dirichlet), because NRBCs and HABCs are BCs themselves

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -1,55 +1,106 @@
...
+
+    Raises
+    ------
+    ValueError
+        If `c` is given together with `K` or `rho_flui`, or if
+        only one of `K`/`rho_fluid` is given.
     """
-    V = wave.function_space
-    quad_rule = wave.quadrature_rule
+    if c is not None and (K is not None or rho_fluid is not None):
```
ruansava, line 35: Use `mutually_exclusive_parameter_error()` method in `spyro.utils.error_management.py`

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -1,55 +1,106 @@
...
+        raise ValueError(
+            "Ambiguous formulation. Insert 'bulk_modulus' and 'density_fluid' "
```
ruansava, line 37: I believe a method should be implemented to establish the equivalence between properties such as this one: `https://en.wikipedia.org/wiki/P-wave_modulus`

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -1,55 +1,106 @@
...
+    form : UFL form
+        The combined weak form (mass + stiffness + source + absorbing terms).
+
+    Raises
+    ------
+    ValueError
+        If `c` is given together with `K` or `rho_fluid` (ambiguous), or if
+        only one of `K`/`rho_fluid` is given (incomplete).
+    """
+    c = velocity_fluid(c, K, rho_fluid, default=wave.c)
```
ruansava, line 84: I think that this method should depend only on `wave` object. Something like the modifications in PR #373

ruansava (changes requested):

spyro/solvers/elastic_wave/forms.py
Olender: Yes, I agree

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: Done

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: This is intentional because in the coupled case u_n and u_nm1 aren't wave.u_n and wave.u_nm1 directly, they are components extracted from the mixed space using fire.split

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: It's true. Done, thanks.

spyro/solvers/elastic_wave/forms.py
j-salesps: Would just "isotropic_elastic" be better?

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: I just changed because of the "make format" request. See if it's better now in the next commit

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: Done. I saw the need to create a method to display an error message if either the `K` or `rho_fluid` parameter is not provided. I named it `required_together_parameter_error` in `error_management.py`. Please check if it is suitable, as well as my fix in `acoustic_solver_construction_no_pml.py`.

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: I'm not entirely sure where to implement this. Do you mean it can also be used to calculate other parameters such as Vp and Vs from the Lamé parameters?

spyro/solvers/acoustic_solver_construction_no_pml.py
j-salesps: Done. Please, check if that's what you needed.

spyro/solvers/elastic_wave/forms.py
ruansava: For the moment, I think that it can work

spyro/solvers/acoustic_solver_construction_no_pml.py
ruansava: We can extract them "in situ", see PML solver for acoustic case

spyro/utils/error_management.py
```diff
@@ -664,3 +664,39 @@ def _join_options(values, conjunction="or"):
...
+    """Raise a ValueError if some parameters in a group aren't given.
+
+    Parameters
+    ----------
+    parameter_name_lst : `list` of `str`
+        List of names of the parameters that must be given together.
+    parameter_value_lst : `list`
+        List of values of the parameters that must be given together.
+
+    Riases
```
ruansava, line 679: typo: Raises

spyro/utils/error_management.py
```diff
@@ -664,3 +664,39 @@ def _join_options(values, conjunction="or"):
         return values[0]
 
     return f"{', '.join(values[:-1])} {conjunction} {values[-1]}"
+
+
+def required_together_parameter_error(parameter_name_lst, parameter_value_lst):
```
ruansava, line 669: Add tests in `tests.on_one_core.test_error_manager.py`

spyro/solvers/acoustic_solver_construction_no_pml.py
```diff
@@ -1,55 +1,103 @@
...
-    V = wave.function_space
-    quad_rule = wave.quadrature_rule
+    c = wave.c
+    K = wave.K
+    rho_fluid = wave.rho_fluid
 
-    # typical CG FEM in 2d/3d
-    u = fire.TrialFunction(V)
-    v = fire.TestFunction(V)
+    combined_k_and_rho_fluid = K if K is not None else rho_fluid
```
ruansava, line 40: Add validation parameters from `tests.on_one_core.test_error_manager.py`to deal with None values

ruansava (changes requested):

PR #405, opened 2026-09-15
Demo fwi elastic auto adj

Ig-dolci:
## Elastic FWI demo with the automated adjoint

Adds a demo of a 2D isotropic elastic full-waveform inversion driven by the automated adjoint (`firedrake.adjoint`).

### Demo

`demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst` is a reStructuredText document whose `.. code-block:: python` blocks form the program (extracted with `pylit`, run with `mpiexec -n 3`, one process per shot). It inverts jointly for the P- and S-wave velocities of a circular inclusion, with density fixed:

- synthetic data from a true model, a Gaussian starting model, spectral elements (Q4, lumped) with Stacey absorbing boundaries, ensemble parallelism over the shots;
- units made explicit where the parameters are defined (km, s, km/s, g/cm³ ⇒ GPa);
- gradient verification with a seeded random Taylor test (`verify_gradient`, rate 2.000);
- bound-constrained inversion with `LumpedTAOSolver` (PETSc/TAO BLMVM) for 20 iterations, single-memory checkpointing with periodic garbage collection;
- figures of the true, starting and inverted models, the observed records and the functional history, with a short discussion of the result (J drops by a factor of ~21; c_s is recovered well, c_p much less — the circle is one S wavelength wide but half a P wavelength).

`demos/README.md` documents how to extract and run it; `.gitignore` allows the demo's figures.

### Library

- `spyro/tools/optimization.py`: `LumpedTAOSolver` now seeds BLMVM's initial Hessian `H0` with the lumped Riesz map (as pyadjoint's `TAOSolver` does).
- `spyro/solvers/automatic_differentiation_solver.py`: `clear_tape()` gives the controls fresh block variables, so a new recording starts from the controls' *current* values rather than from the checkpoint left by the last reduced-functional evaluation (a Taylor-test perturbation or the last iterate). Regression test in `test_gradient_checkpointing.py`, for all three schedules.
- `spyro/plots/general_plots.py`: `plot_model` generalised — any scalar fields (`fields=`), or the solver's material parameters by default; multi-panel layout with titles/limits per panel; optional CG1 regridding (`high_resolution`) with one mesh per call; drawn with `FunctionPlotter`/`tripcolor` on owned cells only, gathered to rank 0, so it works on distributed meshes; `plot_model_in_p1` is a thin wrapper. `plot_shots` autoscales its colour range when no limits are given. `change_scalar_field_resolution` can reuse a regridding space.

### Tests

- `tests/demo_tests/test_elastic_fwi_automated_adjoint_demo.py`: extracts the demo's code blocks, coarsens the mesh/time step/iterations, and runs it on 3 ranks (added to the newer-Firedrake CI job).
- `tests/on_one_core/test_plots.py`, `tests/parallel/test_plot_model.py`: the generalised `plot_model` (fields, layout, regridding, elastic parameters, distributed mesh).
- `tests/on_one_core/test_gradient_checkpointing.py`: the `clear_tape` fix.

### Follow-up

This PR inverts both velocities at once. Future PRs will add the inversion of the control parameters separately, one parameter at a time, in stages, with the others held.


demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+inclusion in a two-dimensional elastic medium. Density stays fixed. We
+first generate synthetic observations, choose a starting model, check the
+adjoint gradient, and then run the inversion.
+
+Running the demo
+--------------------
+
+Use a spyro environment with Firedrake's adjoint and ``PointEvaluator``
+support (the demo is tested in CI with Firedrake 2026.4.0), PETSc/TAO,
+Matplotlib and `pylit <https://pypi.org/project/pylit/>`__.
```
Ig-dolci:
```suggestion
I recommend using the latest Firedrake release version to run this demo.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
+Elastic full-waveform inversion with the automated adjoint
+==============================================================
+
+This demo uses spyro to recover the P- and S-wave velocities of a circular
+inclusion in a two-dimensional elastic medium. Density stays fixed. We
+first generate synthetic observations, choose a starting model, check the
+adjoint gradient, and then run the inversion.
+
```
Ig-dolci, line 8:
```suggestion

`Daiane I Dolci <https://ig-dolci.github.io/>`__ prepared this tutorial.

```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+three MPI processes::
+
+    pylit --code-block-marker ".. code-block:: python" elastic_fwi_automated_adjoint.py.rst
+    mpiexec -n 3 python elastic_fwi_automated_adjoint.py
+
+The ``.. code-block:: python`` blocks below form the complete program, in
+order. The run saves model plots, shot records as figures, and inversion
+outputs in the current directory.
+
+What is being minimised?
```
Ig-dolci:
```suggestion
Full Waveform Inversion (FWI)
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+----------------------------
+
+Full-waveform inversion (FWI) adjusts material parameters to reduce the
```
Ig-dolci:
```suggestion
FWI seeks to adjust material parameters to reduce the
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+
+For each shot :math:`s`, :math:`\mathbf{u}_s` is the predicted displacement
+and :math:`\mathbf{d}_s` is the observation at receiver :math:`\mathbf{x}_r`.
+The norm includes both displacement components. spyro approximates the time
+integral with the trapezoidal rule on the solver's time grid.
+
+In this synthetic example, the observations come from a forward solve with
+a known *true model*. The inversion begins from a different *starting
+model*. FWI is a local optimisation: a good starting model helps, but a
+smaller misfit does not guarantee recovery of the true medium.
```
Ig-dolci:
```suggestion
model*.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+    \mu = \rho c_s^2, \qquad \lambda = \rho(c_p^2 - 2c_s^2).
+
+The medium starts at rest and is excited by vertical point forces with a
+5 Hz Ricker wavelet [Ricker1953]_. We use degree-4 spectral elements on
+quadrilaterals [Komatitsch1998]_ and central differences in time. Stacey
+absorbing conditions [Stacey1988]_ reduce reflections at all four edges;
+there is no free surface. Sources and receivers lie on opposite sides of
+the inclusion, forming a transmission experiment.
+
+How the automated adjoint fits in
```
Ig-dolci:
```suggestion
Automated Adjoints
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+
+``firedrake.adjoint`` records the forward computation on a *tape* and
+traverses it backwards to differentiate the discrete misfit. A *reduced
+functional* wraps that computation so the optimiser can evaluate the misfit
+and its gradient at new control values.
+
+The demo uses ``SingleMemoryStorageSchedule``: all states needed by the
+adjoint stay in memory, with no forward recomputation. For larger problems,
+setting ``snapshots`` enables a schedule that stores fewer checkpoints and
+recomputes intermediate states [Dolci2024]_.
```
Ig-dolci:
```suggestion
recomputes intermediate states employing the Mixed Schedule checkpointing strategy [Maddison2024]_.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+        "frequency": frequency,
+        "amplitude": np.array([1.0, 0.0]),   # vertical unit force (z, x)
+        "receiver_locations": spyro.create_transect(
+            (-0.85, 0.1), (-0.85, 0.9), 41,
+        ),
+    }
+
+``synthetic_data`` supplies the background material. spyro computes
+:math:`\lambda` and :math:`\mu` from these values; we add the true and
+starting anomalies below.
```
Ig-dolci:
```suggestion
:math:`\lambda` and :math:`\mu` from these values. The true and starting-velocity models are shown below.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+
+.. image:: elastic_fwi_observed_ux[0].png
+    :width: 45 %
+    :alt: observed horizontal displacement record of the first shot
+
+The starting model and the controls
+---------------------------------------
+
+The inversion runs on the *guess* mesh, built here with the same element
+size as the mesh the data were generated on.
```
Ig-dolci:
```suggestion
For the sake of simplicity, the inversion runs on the *guess* mesh with the same element
size as the mesh on which the observed data were generated.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+
+.. image:: elastic_fwi_starting_model.png
+    :width: 90 %
+    :alt: starting P- and S-wave velocity models
+    :align: center
+
+Enable the automated adjoint for the two velocity fields. Selecting these
+*controls* leaves density fixed. ``checkpointing=True`` uses the memory
+schedule described above; periodic garbage collection releases reference
+cycles, but does not replace checkpointing.
```
Ig-dolci:
```suggestion
*controls* leaves density fixed. ``checkpointing=True`` uses a single memory
schedule that does not recompute the forward solver and stores only the forward data used for the adjoint-based gradient. Periodic garbage collection releases reference cycles. This is a good approach once the code architecture used to obtain the automated gradient is passive of reference cycles.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+.. math::
+
+    \left|J(m+h\delta m)-J(m)-h\,DJ(m)[\delta m]\right| = O(h^2).
+
+Thus, halving :math:`h` should reduce the residual by about four, until
+round-off dominates. ``verify_gradient`` pairs the gradient fields with the
+perturbations and returns the observed convergence rate, which should be
+close to two. We use the same random seed on every ensemble member so their
+perturbations agree. The test uses the unmasked gradient; the mask below
+only restricts optimisation updates.
```
Ig-dolci:
```suggestion
perturbations agree.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+the largest allowed P-wave speed, rather than just the starting velocity.
+
+.. code-block:: python
+
+    cp_bounds = (2.3, 3.5)   # km/s
+    cs_bounds = (1.0, 1.6)   # km/s
+
+Large gradients near sources and receivers can dominate model updates
+[Modrak2016]_. This binary mask allows updates only in the strip
+:math:`-0.75 < z < -0.3` km, holding the starting model fixed elsewhere.
```
Ig-dolci:
```suggestion
Large gradients near sources and receivers can dominate model updates
[Modrak2016]_. A common practice in this case is to zero out such regions. This binary mask allows updates only in the strip:
math:`-0.75 < z < -0.3` km, holding the starting model fixed elsewhere.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,612 @@
...
+
+.. image:: elastic_fwi_functional_history.png
+    :width: 50 %
+    :alt: misfit functional against iteration
+    :align: center
+
+
+A falling misfit means the predicted records better match the observations.
+Compare both velocity maps as well: the two parameters need not improve at
+the same rate, and a good data fit can still leave model errors.
```
Ig-dolci:
```suggestion
A decreasing misfit means the predicted records better match the observations. We can achieve a better predicted model via FWI using more iterations, or other strategies for multiparameter inversion problems.
It is an exercise for you!
```

Ig-dolci (commented):
Apply changes

spyro/plots/general_plots.py
```diff
@@ -335,3 +345,199 @@ def plot_function(function: Function, **kwargs) -> None:
...
+    z_max, x_max = (mesh.comm.allreduce(float(v), op=MPI.MAX) for v in local_max)
+    # Include both edges without rounding the last sample outside the mesh.
+    depths = np.linspace(z_max, z_min, max(2, int(np.ceil((z_max - z_min) / spacing)) + 1))
+    positions = np.linspace(x_min, x_max, max(2, int(np.ceil((x_max - x_min) / spacing)) + 1))
+    Z, X = np.meshgrid(depths, positions, indexing="ij")
+    points = np.column_stack([Z.ravel(), X.ravel()])
+    return points, (len(depths), len(positions), x_min, x_max, z_min, z_max)
+
+
+def plot_scalar_field(
```
Ig-dolci, line 404: See plot_model and check for high resolution that is there.

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,608 @@
...
+    vs_circle = 1.5        # km/s
+
+    center_z, center_x = -0.5, 0.5   # centre of the circle, km
+    radius = 0.125                   # km
+
+The domain is a 1 km × 1 km square. spyro's first coordinate is the depth
+:math:`z`, which is zero at the top and *negative* below it, and the second
+is the horizontal position :math:`x`. The mesh is a uniform grid of
+quadrilaterals of 100 m (``edge_length``), and the Ricker wavelet has a peak
+frequency of 5 Hz. The time step of 1.6 ms is close to the stability limit
```
Olender, line 141: This might not be the ebst place for this discussion, but I've been running some CPW calcualtions for elastic. Your choice of edge length represents 2.5 cells-per-wavelength if we were to use the Ricker peak frequency (fp) as a reference (which is what I was doing based on Lyu's 2020 paper). However, we should start following Mondaic, Fitchners papers, SEPCFEM, and Lyus 2024 paper were the cpw reference frequency is fmax = 2*fp (because most of the source energy is below this line). In that case, this demo is using 1.25 cpw, which might be worth a mention right beside the edge length.

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,608 @@
...
+        f"{comm.ensemble_comm.size} ensemble member(s), "
+        f"{comm.comm.size} process(es) each", comm,
+    )
+
+The true model and the observed data
+----------------------------------------
+
+``set_real_mesh`` builds the mesh for generating observations. A smoothed
+indicator, close to one inside the circle and zero outside, defines the
+velocity anomalies. The hyperbolic tangent smooths their edges.
```
Olender, line 287: Maybe mention that ebcause you are using a different mesh for the true model you are also avoiding inverse crime?

Olender: My apologies, I just noticed you are using the same mesh for both. What do you think about changing the mesh?

spyro/plots/general_plots.py
```diff
@@ -283,6 +404,10 @@ def plot_shots(
     else:
         arr = wave.forward_solution_receivers[:, :, out_index]
 
+    if vmin is None and vmax is None:
```
Olender, line 407: Gostei

spyro/tools/optimization.py
```diff
@@ -198,11 +198,61 @@ def _lumped_riesz_map(controls, comm=None):
...
+        matrix.
+    comm : petsc4py.PETSc.Comm or mpi4py.MPI.Comm
+        Communicator the controls are defined over.
+
+    Returns
+    -------
+    tuple of (petsc4py.PETSc.Mat, petsc4py.PETSc.PC)
+        The initial Hessian, and the preconditioner applying its inverse.
+    """
+    from petsc4py import PETSc
```
Olender, line 224: Why import here isntead of at the top?

spyro/tools/optimization.py
```diff
@@ -211,22 +261,24 @@ class LumpedTAOSolver(OptimizationSolver):
...
-        starting point, even where both converge.
+    The initial Hessian
+        BLMVM builds its inverse-Hessian approximation from an ``H0``, which
+        also sets the first step, along :math:`-H_0^{-1} DJ`. Like pyadjoint,
+        this seeds ``H0`` with the Riesz map -- the lumped one -- so the
+        first step follows the gradient in the same metric as the
+        convergence test. PETSc's default, a scaled identity, would step
+        along the raw coefficient derivative; on spectral elements, where
+        the nodal masses vary by nearly two orders of magnitude within an
+        element, that moves the interior nodes and leaves the edges behind.
```
Olender, line 272: Que legal!

spyro/utils/velocity_to_grid.py
```diff
@@ -106,6 +111,9 @@ def change_scalar_field_resolution(scalar_field, mesh_parameters, grid_spacing):
         Mesh metadata describing the original model domain.
     grid_spacing : float
         Desired grid spacing (edge length) for the new structured mesh (km).
+    function_space : firedrake.FunctionSpace, optional
```
Olender, line 114: Boa!

Olender (approved):

Olender:
Fiz uns poucos comentários na demo, mas ela está boa e fica a seu critério se vale a pena mudar ou não

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
Ig-dolci: I tried to keep this demo as simple as possible. However, it is interesting to add a note at the end with an exercise on changing the mesh.

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,608 @@
...
+    :align: center
+
+
+After 20 iterations the misfit has fallen by a factor of about 21. The
+S-wave velocity has become a circle of the right size and amplitude, about
+1.48 km/s at the centre against a true 1.5, while the P-wave velocity has
+moved much less, from 2.75 to about 2.79 km/s against 3.0. Both models
+also carry an imprint of the acquisition: spots at the sources and ripples
+along the line of receivers, where the wavefields, and so the gradient,
+are strongest [Modrak2016]_.
```
Ig-dolci:
```suggestion
moved much less, from 2.75 to about 2.79 km/s against 3.0.
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,608 @@
...
+    America, 88(2), 368–392.
+
+.. [Maddison2024] Maddison, J. R. (2024). Step-based checkpointing with
+    high-level algorithmic differentiation. Journal of Computational
+    Science, 82, 102405.
+
+.. [Modrak2016] Modrak, R., & Tromp, J. (2016). Seismic waveform inversion
+    best practices: regional, global and exploration test cases.
+    Geophysical Journal International, 206(3), 1864–1889.
+
```
Ig-dolci:
```suggestion
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
Ig-dolci: I just realised that it is not a good idea to add this exercise yet. We first need to verify whether Spyro can compute the misfit when the observed and predicted receiver data use different time-step sizes. For example, the observed data might be generated with \(dt = 0.0005\), while the predicted data are computed with \(dt = 0.0075\). In that case, we need a proper way to align or interpolate the time samples before evaluating the misfit.

Because using different meshes in this demo may require different stable time-step sizes, the exercise would raise this unresolved issue. :(

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,608 @@
...
+of this mesh for the fastest velocity the inversion is allowed to reach,
+and the records are 1 s long: enough for the P wave, which arrives from
+about 0.5 s, and the slower S wave, from about 0.8 s.
+
+.. code-block:: python
+
+    length_z = 1.0   # depth of the domain, km
+    length_x = 1.0   # width of the domain, km
+
+    edge_length = 0.1     # element size, km
```
Ig-dolci:
```suggestion
    edge_length = 0.1  # km; 1.25 cells per shortest wavelength at fmax = 2*fp
```

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
Ig-dolci: I am going to add a comment on `edge_lengh` right side.

demos/elastic_fwi_automated_adjoint/elastic_fwi_automated_adjoint.py.rst
```diff
@@ -0,0 +1,608 @@
...
+    radius = 0.125                   # km
+
+The domain is a 1 km × 1 km square. spyro's first coordinate is the depth
+:math:`z`, which is zero at the top and *negative* below it, and the second
+is the horizontal position :math:`x`. The mesh is a uniform grid of
+quadrilaterals of 100 m (``edge_length``), and the Ricker wavelet has a peak
+frequency of 5 Hz. The time step of 1.6 ms is close to the stability limit
+of this mesh for the fastest velocity the inversion is allowed to reach,
+and the records are 1 s long: enough for the P wave, which arrives from
+about 0.5 s, and the slower S wave, from about 0.8 s.
```
Ig-dolci:
```suggestion
and the records are 1 s long. The direct-wave arrival times can be estimated
from a vertically aligned source--receiver pair, whose propagation distance is

.. math::

    d = \lvert -0.85 - (-0.15) \rvert = 0.70\ \mathrm{km}.

Using the background velocities, the corresponding travel times are

.. math::

    t_P^{\mathrm{travel}} = \frac{d}{v_P}
    = \frac{0.70}{2.5} = 0.28\ \mathrm{s},
    \qquad
    t_S^{\mathrm{travel}} = \frac{d}{v_S}
    = \frac{0.70}{1.25} = 0.56\ \mathrm{s}.

Spyro's default Ricker-wavelet delay places its peak at

.. math::

    t_0 = \frac{1.5\sqrt{6}}{\pi f_p} \simeq 0.234\ \mathrm{s}

for :math:`f_p = 5` Hz. The peaks of the direct arrivals are therefore
expected at approximately :math:`t_0 + t_P^{\mathrm{travel}} = 0.51` s and
:math:`t_0 + t_S^{\mathrm{travel}} = 0.79` s. A final time of 1 s captures
both arrivals at the vertically aligned receivers; receivers with a larger
horizontal offset have later arrivals.
```

Ig-dolci (commented):
Review

